PR #1, opened 2021-07-06
dlc

nickgreenspan:
Added dlc-like model and accompanying dataset, training script, and tests to repo.


danbider (approved):
Looks good, pull request accepted but we should generalize some parts of the code that are specific to our current 384X384 dataset if possible.

PR #23, opened 2021-11-11
subpixel maxima update

themattinthehatt:
Removed SubPixMaxima class transcribed from DPK, which contained an argmax that pytorch could not pass gradients through - this affected unsupervised losses like temporal and pca. Replaced this with a fully differentiable soft argmax function from the kornia library, which is now a new (lightweight) dependency.


pose_est_nets/models/heatmap_tracker.py
```diff
@@ -214,7 +227,7 @@ def heatmaps_from_representations(
     @typechecked
     def forward(
         self, images: TensorType["batch", "channels":3, "image_height", "image_width"]
```
danbider, line 229: if we're adding `float`s (which I hope won't hurt when we try half precision), I'd suggest adding one for `images`.

tests/test_subpixmaxima.py
```diff
@@ -10,145 +12,113 @@
...
-        device=_TORCH_DEVICE,
-    )
     example = dataset.__getitem__(idx=0)
-    maxima, confidence = SubPixMax.run(
-        example["heatmaps"].unsqueeze(0).to(_TORCH_DEVICE)
+    assert example["keypoints"].shape == (torch.Size([34]))
+    heatmap_gt = example["heatmaps"].unsqueeze(0)
+    keypts_gt = example["keypoints"].unsqueeze(0).reshape(1, 17, 2)
+    assert heatmap_gt.shape == (1, 17, 96, 96)
+    heatmap_torch = generate_heatmaps(
```
danbider, line 31: do we mind replacing it with `kornia`'s `render_gaussian_2d`? it'll lighten our repo and be a more tested version of `generate_heatmaps`. Another plus is that the heatmap is generated and with the same machinery that then does the subpixmax ops.
https://kornia.readthedocs.io/en/v0.2.0/_modules/kornia/geometry/dsnt.html#render_gaussian_2d

tests/test_subpixmaxima.py
```diff
@@ -10,145 +12,113 @@
...
-        coordinate_scale=torch.tensor(4, device=_TORCH_DEVICE),  # 2 ** 2
-        confidence_scale=torch.tensor(1, device=_TORCH_DEVICE),
-        threshold=None,
-        device=_TORCH_DEVICE,
-    )
-    og_maxima, gt_confidence = SubPixMax.run(gt_heatmap.to(_TORCH_DEVICE))
-    torch_maxima, confidence_t = SubPixMax.run(torch_heatmap.to(_TORCH_DEVICE))
-    assert (og_maxima == torch_maxima).all()
-    assert (gt_confidence == confidence_t).all()
+    assert(preds_gt == preds_torch).all()
```
danbider, line 49: seems like it could be a strong test, but if I understand correctly `example_heatmaps` and `heatmap_torch` are by definition the same thing. so we're not really testing the subpix maxima no? 
the stronger test is to compare the 2D gt keypoints and the 2d predicted keypoints after subpixel maxima. they may not be really equal, but we can accept some `tolerance`. 
Using `render_2d_gaussian` from `kornia` will get you closer to the the equality I think

tests/test_subpixmaxima.py
```diff
@@ -157,22 +127,47 @@ def test_generate_keypoints_batched():
...
-        coordinate_scale=torch.tensor(4, device=_TORCH_DEVICE),  # 2 ** 2
-        confidence_scale=torch.tensor(1, device=_TORCH_DEVICE),
-        threshold=None,
-        device=_TORCH_DEVICE,
-    )
-    for batch in dl:
-        gt_keypts = batch["keypoints"].reshape(-1, 17, 2)
-        heatmaps = generate_heatmaps(
-            gt_keypts, height=384, width=384, output_shape=(96, 96)
+    for batch_dict in dl:
```
danbider, line 130: again not sure what we're testing here?

tests/test_subpixmaxima.py
```diff
@@ -157,22 +127,47 @@ def test_generate_keypoints_batched():
...
-        torch_maxima, confidence_t = SubPixMax.run(heatmaps.to(_TORCH_DEVICE))
-        assert (gt_maxima == torch_maxima).all()
-        loss = MaskedMSEHeatmapLoss(batch["heatmaps"], heatmaps)
+        confidences_torch = torch.amax(softmaxes_torch, dim=(2, 3))
+
+        # og_maxima, gt_confidence = SubPixMax.run(heatmap_gt.to(_TORCH_DEVICE))
+        # torch_maxima, confidence_t = SubPixMax.run(heatmap_torch.to(_TORCH_DEVICE))
+        print(preds_gt[-1])
+        print(preds_torch[-1])
+        assert torch.allclose(preds_gt, preds_torch)
```
danbider, line 163: that's a cool one! happy to learn it from you

danbider (changes requested):
- typechecking suggestions
- a question on `test_subpixmaxima.py`
- suggestion to add `kornia`s `render_2d_gaussian`

pose_est_nets/models/heatmap_tracker.py
```diff
@@ -121,20 +119,35 @@ def num_filters_for_upsampling(self):
...
+    @typechecked
+    def run_subpixelmaxima(
+        self,
+        heatmaps: TensorType[
+            "batch",
+            "num_keypoints",
+            "heatmap_height",
+            "heatmap_width",
+            float]
+    ) -> Tuple[torch.tensor, torch.tensor]:
```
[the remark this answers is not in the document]
themattinthehatt: i'll hold off on the dict update for now but TensorType is a good idea

tests/test_subpixmaxima.py
themattinthehatt: yep that's a good idea

tests/test_subpixmaxima.py
themattinthehatt: if i understand correctly this test is actually for `generate_heatmaps` (despite its name) so we can just scrap altogether if we replace with kornia functions

tests/test_subpixmaxima.py
themattinthehatt: agreed, i just auto-replaced everything; ideally we could delete this entire test module eventually

tests/test_subpixmaxima.py
themattinthehatt: :)

danbider (approved):
great

PR #25, opened 2021-11-13
Feature/evaluate

nickgreenspan:
Added automatic creation of model prediction csv files after training. Refactored Matt's video prediction script to work on a labeled dataset in addition to videos. Added tests for the prediction scripts and uploaded a toy config and a toy model path to do so.

### DIFF
--- a/pose_est_nets/utils/plotting_utils.py
+++ b/pose_est_nets/utils/plotting_utils.py
@@ -105,7 +105,7 @@ def predict_videos(
     cfg_file: Union[str, DictConfig],
     save_file: Optional[str] = None,
     sequence_length: int = 16,
-    device: Literal["gpu", "cuda", "cpu"] = "gpu", #TODO: can we just use _TORCH_DEVICE here?
+    device: Literal["gpu", "cuda", "cpu"] = "gpu",
     video_pipe_kwargs={},
 ):
     """Loop over a list of videos and process with tracker using DALI for fast inference.
@@ -378,11 +378,12 @@ def predict_dataset(
     datamod: LightningDataModule, 
     hydra_output_directory: str,
     ckpt_file: str,
+    save_file: str = None,
     heatmap_idxs: List[int] = None
 ):
     """
     Call this function with a path to ckpt file for a trained model
-    heatmap_idxs: indexes of datapoints to save heatmaps for
+    heatmap_idxs: indexes of datapoints to save heatmaps for (NOT IMPLEMENTED YET)
 
     """
     model = load_model_from_checkpoint(cfg=cfg, ckpt_file=ckpt_file, eval=True)
@@ -398,7 +399,8 @@ def predict_dataset(
         json.dump(dataset_split_indices, f)
 
     full_dataloader = DataLoader(dataset = full_dataset, batch_size = datamod.test_batch_size)
-    save_file = hydra_output_directory + "/predictions.csv" #default for now, should be saved to the model directory
+    if save_file is None: 
+       save_file = hydra_output_directory + "/predictions.csv" #default for now, should be saved to the model directory
     make_predictions_and_create_csv(
         cfg = cfg, 
         model = model, 
--- a/pose_est_nets/utils/plotting_utils.py
+++ b/pose_est_nets/utils/plotting_utils.py
@@ -21,7 +21,7 @@ import os
 import csv
 import json
 import time
-from typing import Callable, Optional, Tuple, List, Union, Literal
+from typing import Callable, Optional, Tuple, List, Union, Literal, Type
 from typeguard import typechecked
 from tqdm import tqdm
 from omegaconf import DictConfig, OmegaConf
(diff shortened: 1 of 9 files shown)


pose_est_nets/utils/plotting_utils.py
```diff
@@ -351,94 +257,153 @@ def predict_videos(
...
+    cfg: DictConfig,
+    model: LightningModule,
+    dataloader: torch.utils.data.DataLoader,
+    n_frames_: int, #total number of frames,
+    batch_size: int, #batch size or sequence length
+    save_file: str,  
+    data_name: str = "dataset",
+    save_heatmaps: bool = False,
+):
+    keypoints_np = np.zeros((n_frames_, model.num_keypoints * 2))
```
danbider, line 286: great refactoring. it is still a long function that can be further nicely refactored into three (still large) funcs -- 
1. `predict()`: lines `286-328`
2. `undo_resize()`: lines `330-342`
3.  `build_df_and_save()`: lines `344-373`

pose_est_nets/utils/plotting_utils.py
```diff
@@ -351,94 +257,153 @@ def predict_videos(
...
+    dataset_split_indices = {
+        "train": datamod.train_dataset.indices,
+        "validation": datamod.val_dataset.indices,
+        "test": datamod.test_dataset.indices,
+    }
+    with open(hydra_output_directory + "/dataset_split_indices.json", "w") as f:
+        json.dump(dataset_split_indices, f)
+
+    full_dataloader = DataLoader(dataset = full_dataset, batch_size = datamod.test_batch_size)
+    save_file = hydra_output_directory + "/predictions.csv" #default for now, should be saved to the model directory
```
danbider, line 401: `save_file` should be made an input to the function, with default `None`. then, before line 401, add 
```
if save_file is None: 
       save_file = hydra_output_directory + "/predictions.csv"
```

toy_datasets/toy_model/config.yaml
```diff
@@ -0,0 +1,94 @@
+training:
```
danbider, line 1: should be deleted from the `toy_datasets/toy_model` directory

pose_est_nets/utils/plotting_utils.py
```diff
@@ -351,94 +257,153 @@ def predict_videos(
...
+    """
+    Call this function with a path to ckpt file for a trained model
+    heatmap_idxs: indexes of datapoints to save heatmaps for
+
+    """
+    model = load_model_from_checkpoint(cfg=cfg, ckpt_file=ckpt_file, eval=True)
+    model.to(_TORCH_DEVICE)
+    full_dataset = datamod.dataset
+    num_datapoints = len(full_dataset)
+    dataset_split_indices = {
```
danbider, line 392: we should consider adding these jsons to the dataframe which we're saving. add a column called `set`, and each row is either `train`/`val`/`test`.

danbider (changes requested):
Changes that are very much needed. Comments:

- the `make_predictions_and_create_csv()` function is still quite large, and can be broken down to its three major parts (see detailed suggestion in the review).
- I think that instead of saving a separate `.json` with `train/test/val`, we can add this info into a new column in the dataframe which we're saving. (appending a column to an existing dataframe is very easy). it'll also make it super easy for the FiftyOne function. we can just loop over the rows of this dataframe and have all the information needed.
- better to delete all test related configs and checkpoints from toydataset. adding many MBs to repos makes them harder to download.

pose_est_nets/utils/plotting_utils.py
```diff
@@ -351,94 +257,153 @@ def predict_videos(
...
     # if iterating over multiple models, outside this function, the below will reduce
     # memory
     del model, pipe, predict_loader
     torch.cuda.empty_cache()
+
+def make_predictions_and_create_csv(
+    cfg: DictConfig,
+    model: LightningModule,
+    dataloader: torch.utils.data.DataLoader,
+    n_frames_: int, #total number of frames,
```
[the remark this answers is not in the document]
themattinthehatt: is `n_frames` necessary as an input here? can we get this directly from the dataloader?

pose_est_nets/utils/plotting_utils.py
nickgreenspan: I agree that makes sense. The dataset prediction csvs would then have a different format than the video prediction csvs but I guess that's fine.

pose_est_nets/utils/plotting_utils.py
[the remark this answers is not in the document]
danbider: `n_frames` is extracted differently for `LightningWrapper` and standard `DataLoader`. could add a condition to test these differently, but we can also keep this as is for a little while

pose_est_nets/utils/plotting_utils.py
danbider: made 1,2. still need to do 3

PR #27, opened 2021-12-06
Feature/heatmap losses

nickgreenspan:
Added wasserstein heatmap loss support, added script to make fiftyone visualization from csvs with heatmaps, changed ground truth heatmap creation to pytorch generate_heatmaps function, added support for uniform heatmaps there too.

### DIFF
--- a/pose_est_nets/utils/plotting_utils.py
+++ b/pose_est_nets/utils/plotting_utils.py
@@ -314,25 +314,6 @@ def predict_frames(
                 image = batch  # predicting from video
             outputs = model.forward(image)
             if cfg.model.model_type == "heatmap":
-                # if type(batch) == dict and save_heatmaps: #only predicting heatmaps for dataset for now
-                    #plt.imshow(batch["images"][0][kpt_plt_idx].numpy())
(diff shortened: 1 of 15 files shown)


pose_est_nets/utils/plotting_utils.py
```diff
@@ -445,7 +474,7 @@ def make_predictions(
     # build dataframe from the hierarchal index and predictions array
     df = pd.DataFrame(predictions, columns=pd_index)
 
-    return df
+    return df, heatmaps_np
```
danbider, line 477: how does returning `heatmaps_np` play out with the regression model? did you try running this func? just so it doesn't break things?

danbider (approved):

pose_est_nets/utils/plotting_utils.py
nickgreenspan: I think it should work fine for regression tracker, we set the heatmaps_np to none if its regression model and we only try to save the heatmaps if they are not none. I can test this later when I start up an instance to do some other stuff.

PR #41, opened 2022-06-14
small updates from 06/14/2022

themattinthehatt:
* export frame filenames with prediction csvs
* incorporate new prediction machinery into predict_dataset

### DIFF
--- a/lightning_pose/data/datasets.py
+++ b/lightning_pose/data/datasets.py
@@ -96,13 +96,17 @@ class BaseTrackingDataset(torch.utils.data.Dataset):
                     raise FileNotFoundError("Could not find csv file!")
                 csv_file = options[0]
 
-        csv_data = pd.read_csv(csv_file, header=header_rows)
-        self.keypoint_names = csv_data.columns.levels[0][1:]  # unused inside class
+        csv_data = pd.read_csv(csv_file, header=header_rows, index_col=0)
+        if header_rows == [1, 2] or header_rows == [0, 1]:
+            # self.keypoint_names = csv_data.columns.levels[0]
+            # ^this returns a sorted list for some reason, don't want that
+            self.keypoint_names = [b[0] for b in csv_data.columns if b[1] == 'x']
+        elif header_rows == [0, 1, 2]:
+            # self.keypoint_names = csv_data.columns.levels[1]
+            self.keypoint_names = [b[1] for b in csv_data.columns if b[2] == 'x']
 
-        self.image_names = list(csv_data.iloc[:, 0])
-        self.keypoints = torch.tensor(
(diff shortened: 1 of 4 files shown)


lightning_pose/data/datasets.py
```diff
@@ -96,13 +96,17 @@ def __init__(
...
+        if header_rows == [1, 2] or header_rows == [0, 1]:
+            # self.keypoint_names = csv_data.columns.levels[0]
+            # ^this returns a sorted list for some reason, don't want that
+            self.keypoint_names = [b[0] for b in csv_data.columns if b[1] == 'x']
+        elif header_rows == [0, 1, 2]:
+            # self.keypoint_names = csv_data.columns.levels[1]
+            self.keypoint_names = [b[1] for b in csv_data.columns if b[2] == 'x']
+
+        self.image_names = list(csv_data.index)
+        self.keypoints = torch.tensor(csv_data.to_numpy(), dtype=torch.float32)
```
danbider, line 109: more elegant

lightning_pose/data/datasets.py
```diff
@@ -96,13 +96,17 @@ def __init__(
...
                 csv_file = options[0]
 
-        csv_data = pd.read_csv(csv_file, header=header_rows)
-        self.keypoint_names = csv_data.columns.levels[0][1:]  # unused inside class
-
-        self.image_names = list(csv_data.iloc[:, 0])
-        self.keypoints = torch.tensor(
-            csv_data.iloc[:, 1:].to_numpy(), dtype=torch.float32
-        )
+        csv_data = pd.read_csv(csv_file, header=header_rows, index_col=0)
```
danbider, line 99: great update.
we do this in other places, including in a `get_keypoint_names()` function in [the predictions file](https://github.com/danbider/lightning-pose/blob/23dc5f22e4b40fa8b71193322f11fca703fd8ec9/lightning_pose/utils/predictions.py#L486). 
better to maybe unify to one functions in the utils.io and call it here and there too.
are you also sure about this way of reading colnames?

lightning_pose/utils/predictions.py
```diff
@@ -74,66 +75,39 @@ def predict_dataset(
...
-    }
-
-    full_dataloader = DataLoader(
-        dataset=full_dataset, batch_size=data_module.test_batch_size
-    )
 
-    df, heatmaps_np = _make_predictions(
-        cfg=cfg,
+    trainer = pl.Trainer(gpus=[gpu_id])
+    labeled_preds = trainer.predict(
```
danbider, line 100: nice

danbider (changes requested):
looks great sorry for the delay. mostly look at my suggestion of reusing a single `get_keypoint_names()` function.

danbider (approved):

PR #52, opened 2022-07-22
Confidence

colehurwitz:
Naive implementation of confidence over 2*sigma pixels. On edges, it will repeat edge values rather than exceed image limits.

### DIFF
--- a/lightning_pose/data/utils.py
+++ b/lightning_pose/data/utils.py
@@ -174,7 +174,8 @@ def count_frames(video_list: Union[List[str], str]) -> int:
 
 @typechecked
 def compute_num_train_frames(
-    len_train_dataset: int, train_frames: Optional[Union[int, float]] = None,
+    len_train_dataset: int,
+    train_frames: Optional[Union[int, float]] = None,
 ) -> int:
     """Quickly compute number of training frames for a given dataset.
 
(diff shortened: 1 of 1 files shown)


colehurwitz:
Leaving the old method for confidence commented out for now

colehurwitz:
Does this logic make sense?

lightning_pose/data/utils.py
```diff
@@ -284,15 +284,40 @@ def generate_heatmaps(
...
-    k = locs[:, :, None, 1, None].type(torch.int64)  # y first
-    l = locs[:, :, 0, None, None].type(torch.int64)  # x second
-    vals = heatmaps[i, j, k, l].squeeze(-1).squeeze(-1)  # get rid of singleton dims
-    return vals
+    k = locs[:, :, None, 1, None].type(torch.int64)
+    l = locs[:, :, 0, None, None].type(torch.int64)
+    pix_to_consider = int(np.ceil(sigma * 2.0)) #get all pixels within two stds.
+    offsets = list(np.arange(-pix_to_consider, pix_to_consider+1))
+    vals_all = []
+    for offset in offsets:
```
danbider, line 314: looks good, height and width may differ (i.e., we may have a non square heatmap). so the max clamping should be `heatmap_height-1` then `heatmap_width-1`. you currently use `max=heatmaps.shape[2]-1` for both. other than that, fine

danbider (changes requested):
can also leave a bit more of an explanation in comments

lightning_pose/data/utils.py
colehurwitz: Fixed, thanks!

colehurwitz:
I think my formatter kicked in.. sorry

danbider (approved):
great

PR #62, opened 2022-10-17
data aug

themattinthehatt:


lightning_pose/data/dali.py
```diff
@@ -82,10 +83,30 @@ def video_pipe(
...
+        brightness = fn.random.uniform(range=(0.75, 1.25))
+        video = fn.brightness_contrast(video, brightness=brightness, contrast=contrast)
+        # # shot noise
+        factor = fn.random.uniform(range=(0.0, 10.0))
+        video = fn.noise.shot(video, factor=factor)
+        # jpeg compression
+        # quality = fn.random.uniform(range=(50, 100), dtype=types.INT32)
+        # video = fn.jpeg_compression_distortion(video, quality=quality)
+    else:
+        matrix = -1
```
danbider, line 109: can you add comment why -1?

lightning_pose/data/dali.py
```diff
@@ -168,72 +145,56 @@ def __init__(
...
+                # shape (1,)
+                transform = out[0]["transform"]
             else:
-                # train with context. pipeline is like for "base" model. but we reshape images.
-                # assume batch_size=1
+                # train with context. pipeline is like for "base" model, but we reshape images
+                # further down. assume batch_size=1
                 if self.context_sequences_successive:
-                    # same as for base model
-                    raw_out = torch.tensor(
```
danbider, line 199: I see what you're doing below. we don't need that `torch.tensor(..., dtype=torch.float)` anymore?

lightning_pose/data/datamodules.py
```diff
@@ -171,6 +171,7 @@ def __init__(
         test_probability: Optional[float] = None,
         train_frames: Optional[float] = None,
         torch_seed: int = 42,
+        imgaug: str = "default",
```
danbider, line 174: maybe `Literal["default", "dlc"] = "default"`?

danbider (changes requested):
just commenting typechecking, basically ready to merge but maybe 5 mins of cleanup will make it better?

lightning_pose/data/dali.py
themattinthehatt: I don't think so...the variable `out` is already a tensor, and we were getting warnings about how that tensor was being accessed. This is how I've been training the latest models, don't think there's a problem.

lightning_pose/models/base.py
```diff
@@ -468,16 +473,19 @@ def get_loss_inputs_unlabeled(self, batch: torch.Tensor) -> dict:
...
+                TensorType["seq_len", "context":5, "RGB":3, "image_height", "image_width", float],
+            ],
+            Union[TensorType["seq_len", 2, 3], TensorType[2, 3], TensorType[1]],
+        ],
         stage: Optional[Literal["train", "val", "test"]] = None,
         anneal_weight: Union[float, torch.Tensor] = 1.0,
     ) -> TensorType[(), float]:
         """Compute and log the losses on a batch of unlabeled data (frames only)."""
-
+        print(batch[0].shape)
```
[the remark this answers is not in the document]
themattinthehatt: good catch

lightning_pose/models/heatmap_tracker.py
```diff
@@ -231,7 +227,8 @@ def forward(
         # softmax temp stays 1 here; to modify for model predictions, see constructor
         return spatial_softmax2d(heatmaps, temperature=torch.tensor([1.0]))
 
-    def get_loss_inputs_labeled(self, batch_dict: HeatmapBatchDict) -> dict:
+    # def get_loss_inputs_labeled(self, batch_dict: HeatmapBatchDict) -> dict:
+    def get_loss_inputs_labeled(self, batch_dict) -> dict:
```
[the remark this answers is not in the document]
themattinthehatt: actually not sure why this was commented, I added the typecheck back in; we have a separate SemiSupervisedHeatmapBatchDict that does just what you suggested

danbider (approved):
great

PR #63, opened 2022-10-18
Misc cleanup + improved typechecking coverage

themattinthehatt:


lightning_pose/data/datasets.py
```diff
@@ -214,6 +194,8 @@ def __getitem__(self, idx: int):
                 transformed_images = np.asarray(images)
                 transformed_keypoints = keypoints_on_image.numpy().reshape(-1)
 
+            # send frames to tensors and normalize
+            # need to loop through because ToTensor transform only operates on single images
             for i, transformed_image in enumerate(transformed_images):
                 transformed_image = self.pytorch_transform(transformed_image)
                 if i == 0:
```
danbider, line 201: I'm not sure why we can't take out 
```
image_frames_tensor = torch.unsqueeze(transformed_image, dim=0)
```
outside of the loop? probably there's a reason

lightning_pose/data/datasets.py
```diff
@@ -223,16 +205,18 @@ def __getitem__(self, idx: int):
...
-            assert transformed_keypoints.shape == (self.num_targets,)
 
-        return {
-            "images": transformed_images,  # shape (3, img_height, img_width) or (5, 3, H, W)
-            "keypoints": torch.from_numpy(transformed_keypoints),  # shape (n_targets,)
-            "idxs": idx,
-        }
+        assert transformed_keypoints.shape == (self.num_targets,)
+
+        return BaseLabeledExampleDict(
```
danbider, line 211: nice

lightning_pose/data/datasets.py
```diff
@@ -119,33 +101,32 @@ def __init__(
...
+    def width(self) -> int:
         # assume resizing transformation is the last imgaug one
         return self.imgaug_transform[-1].get_parameters()[0][1].value
 
     def __len__(self) -> int:
         return len(self.image_names)
 
-    @typechecked
-    def __getitem__(self, idx: int):
+    def __getitem__(self, idx: int) -> BaseLabeledExampleDict:
```
danbider, line 129: I'm wondering if at this stage, where the context model is quite established in the code, we can unify the two methods. instead of a big if else, both methods with have a list of images and keypoints. just in context=false this list will be of length 1, and not 5. 
might make the base code slightly longer, but at least it's all in one controlled chunk.

lightning_pose/data/datasets.py
```diff
@@ -298,13 +270,16 @@ def __init__(
...
     @property
-    def output_shape(self):
+    def output_shape(self) -> tuple:
         return (
             self.height // 2**self.downsample_factor,
             self.width // 2**self.downsample_factor,
         )
 
-    def compute_heatmap(self, example_dict: BaseExampleDict):
+    def compute_heatmap(
```
danbider, line 279: nice

lightning_pose/data/datasets.py
```diff
@@ -298,13 +270,16 @@ def __init__(
...
+        self,
+        example_dict: BaseLabeledExampleDict
+    ) -> TensorType["num_keypoints", "heatmap_height", "heatmap_width"]:
         """Compute 2D heatmaps from arbitrary (x, y) coordinates."""
 
         if self.do_context:
```
danbider, line 285: I think we can eliminate this and use `self.height` and `self.width` properties inherited from `BaseTrackingDataset`, assuming all images in the dataset are resized to be of the same size.

lightning_pose/data/datasets.py
```diff
@@ -350,32 +325,23 @@ def compute_heatmaps(self):
...
+        example_dict: BaseLabeledExampleDict = super().__getitem__(idx)
         if len(self.imgaug_transform) == 1 and isinstance(self.imgaug_transform[0], iaa.Resize):
             # we have a deterministic resizing augmentation; use precomputed heatmaps
             example_dict["heatmaps"] = self.label_heatmaps[idx]
         else:
             # we have a random augmentation; need to recompute heatmaps
             example_dict["heatmaps"] = self.compute_heatmap(example_dict)
         return example_dict
-
-    def get_fully_labeled_idxs(self):  # TODO: make shorter
```
danbider, line 375: we should keep something like that as a util, at least for pca. is this the case?

lightning_pose/data/utils.py
```diff
@@ -6,14 +6,74 @@
...
+    heatmaps: TensorType["batch", "num_keypoints", "heatmap_height", "heatmap_width", float]
+
+
+class UnlabeledBatchDict(TypedDict):
+    """Batch type for unlabeled data."""
+    frames: Union[
+        TensorType["seq_len", "RGB":3, "image_height", "image_width", float],
+        TensorType["seq_len", "context":5, "RGB":3, "image_height", "image_width", float],
+    ]
+    transforms: Union[
```
danbider, line 54: I'd add comments explaining the different cases.... this is very nice!!

lightning_pose/data/utils.py
```diff
@@ -113,9 +172,9 @@ def get_loader(self) -> Union[torch.utils.data.DataLoader, dict]:
         if self.cond == "test":
             return self.data_module.test_dataloader()
 
-    @typechecked
     def verify_labeled_loader(
-        self, loader: Union[torch.utils.data.DataLoader, dict]
+        self,
+        loader: Union[torch.utils.data.DataLoader, dict]
```
danbider, line 177: if we decide to make a typed dict of the supervised+unsupervised dataloaders, we can use it here too. but no rush.

lightning_pose/data/utils.py
```diff
@@ -333,23 +386,17 @@ def generate_heatmaps(
     confidence *= -1
     confidence /= 2 * sigma**2
     confidence = torch.exp(confidence)
-    if not normalize:
```
danbider, line 336: nice

lightning_pose/data/utils.py
```diff
@@ -393,9 +440,9 @@ def evaluate_heatmaps_at_location(
...
 def undo_affine_transform(
-        keypoints: TensorType["seq_len", "n_keypoints", 2],
-        transform: Union[TensorType["seq_len", 2, 3], TensorType[2, 3]]
-) -> TensorType["seq_len", "n_keypoints", 2]:
+    keypoints: TensorType["seq_len", "num_keypoints", 2],
+    transform: Union[TensorType["seq_len", 2, 3], TensorType[2, 3]]
+) -> TensorType["seq_len", "num_keypoints", 2]:
 
     # add 1s to get keypoints in projective geometry coords
     ones = torch.ones(
```
danbider, line 448: If it's ok by you, I prefer not sending things to numpy and cpu, and then back to torch, during training. we can do all the indexing and inverting in torch, no?
1. having this little nondifferentiable block might have a bad effect on gradients.
2. also might slow down training times?

danbider (changes requested):
Very nice. took a while to carefully read and appreciate.
The suggestion I made that is most important to me -- undoing the affine transform in torch on the gpu instead of numpy and cpu.

lightning_pose/data/dali.py
```diff
@@ -151,73 +153,54 @@ def __init__(
...
                 # shape (sequence_length, 3, H, W)
-                images = out[0]["x"]
+                frames = out[0]["frames"]
                 # shape (1,)
-                transform = out[0]["transform"]
+                transforms = out[0]["transforms"]
             else:
                 # train with context. pipeline is like for "base" model, but we reshape images
                 # further down. assume batch_size=1
                 if self.context_sequences_successive:
```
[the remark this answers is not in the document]
themattinthehatt: I'm not quite sure I understand what you mean here - are you talking about having a separate `_dali_output_to_tensors`-like function for each return type?

lightning_pose/data/datasets.py
themattinthehatt: yeah i tried this, unfortunately the pytorch transforms.Normalize() only works on single images for some reason; maybe we can revisit this in the future, seems like we shouldn't have to loop

lightning_pose/data/datasets.py
themattinthehatt: I wondered this too - could be possible, let's maybe discuss after this round of edits

lightning_pose/data/datasets.py
themattinthehatt: yes this is a good idea

lightning_pose/data/utils.py
themattinthehatt: none of the transforms should be differentiable, these are fixed parameters; but i can put everything in torch anyways for consistency

lightning_pose/data/datamodules.py
```diff
@@ -34,12 +35,11 @@ def __init__(
         test_probability: Optional[float] = None,
         train_frames: Optional[Union[float, int]] = None,
         torch_seed: int = 42,
-    ) -> None:
```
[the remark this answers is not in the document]
themattinthehatt: changed

lightning_pose/data/dali.py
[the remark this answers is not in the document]
danbider: https://github.com/danbider/lightning-pose/pull/63#discussion_r1003352848 
e.g., 
```
def _dali_output_to_tensors(num_sequences: Literal["single", "multi"]):
      if num_sequences == "single": 
           frames = out[0]["frames"][0, :, :, :, :] # always batch_size=1
           transforms = out[0]["transforms"][0]
     else: # multiple independent five-frame sequences
           frames = out[0]["frames"]
           transforms = out[0]["transforms"]
    return UnlabeledBatchDict(frames=frames, transforms=transforms)
           ```

danbider (approved):

PR #65, opened 2022-10-25
return heatmaps

themattinthehatt:


lightning_pose/utils/predictions.py
```diff
@@ -128,21 +130,22 @@ def unpack_preds(
...
-        preds_2 = stacked_preds[0:-2] # throw out the last two rows.
+        preds_1 = torch.tile(stacked_preds[0], (2, 1))  # copying twice the prediction for image[2]
+        preds_2 = stacked_preds[0:-2]  # throw out the last two rows.
         preds_combined = torch.vstack([preds_1, preds_2])
         # after concat this has the same length. everything is shifted by two rows.
         # but we don't have valid predictions for the last two elements, so we pad with element -3.
         preds_combined[-2:, :] = preds_combined[-3, :]
 
-        if is_confidence == True:
+        if is_confidence:
```
danbider, line 148: for the `mhcrnn` we don't have to do this, since we are also forming prediction with the single-frame head. agree? this `if` just applies for the simple context model.

danbider (changes requested):
Minor suggestion for not zeroing-out first two and last two frames' confidences with the `mhcrnn` model.
So I guess the strategy was to change the `model.predict_step()` but we don't call `trainer.predict()` when heatmaps are requested?

lightning_pose/utils/predictions.py
```diff
@@ -354,12 +360,31 @@ def predict_single_video(
...
+        keypoints, confidences, heatmaps = _predict_frames(
+            cfg=cfg, model=model, dataloader=predict_loader, n_frames_=pred_handler.frame_count,
+            batch_size=batch_size, return_heatmaps=True)
+        preds = [(torch.tensor(keypoints), torch.tensor(confidences))]
+        if heatmaps is not None:
+            heatmaps_file = preds_file.replace(".csv", "_heatmaps.npy")
+            os.makedirs(os.path.dirname(heatmaps_file), exist_ok=True)
+            np.save(heatmaps_file, heatmaps)
+
+    else:
```
danbider, line 388: please add a comment -- `# vanilla prediction, returns keypoints and their associated confidences.`

danbider (approved):

PR #66, opened 2022-11-08
fix nan handling with zero heatmaps

colehurwitz:
Replace uniform heatmaps with zero heatmaps when labels have NaNs

### DIFF
--- a/lightning_pose/data/utils.py
+++ b/lightning_pose/data/utils.py
@@ -358,7 +358,6 @@ def generate_heatmaps(
     width: int,  # width of full sized image
     output_shape: Tuple[int, int],  # dimensions of downsampled heatmap
     sigma: Union[float, int] = 1.25,  # sigma used for generating heatmaps
-    normalize: bool = True,
 ) -> TensorType["batch", "num_keypoints", "height", "width"]:
     """Generate 2D Gaussian heatmaps from mean and sigma.
 
@@ -380,6 +379,7 @@ def generate_heatmaps(
     keypoints[:, :, 1] *= out_height / height
     keypoints[:, :, 0] *= out_width / width
     nan_idxs = torch.isnan(keypoints)[:, :, 0]
+    nan_idxs = nan_idxs
     xv = torch.arange(out_width, device=keypoints.device)
     yv = torch.arange(out_height, device=keypoints.device)
     # note flipped order because of pytorch's ij and numpy's xy indexing for meshgrid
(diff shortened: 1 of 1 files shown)


colehurwitz:
It passed pytests on my side so I think it should work

lightning_pose/data/utils.py
```diff
@@ -380,6 +379,7 @@ def generate_heatmaps(
     keypoints[:, :, 1] *= out_height / height
     keypoints[:, :, 0] *= out_width / width
     nan_idxs = torch.isnan(keypoints)[:, :, 0]
+    nan_idxs = nan_idxs
```
danbider, line 382: not sure about this line, why is it there? should it be deleted?

lightning_pose/data/utils.py
```diff
@@ -358,7 +358,6 @@ def generate_heatmaps(
     width: int,  # width of full sized image
     output_shape: Tuple[int, int],  # dimensions of downsampled heatmap
     sigma: Union[float, int] = 1.25,  # sigma used for generating heatmaps
-    normalize: bool = True,
```
danbider, line 361: did we agree to remove the normalize argument @themattinthehatt @colehurwitz ?

lightning_pose/data/utils.py
```diff
@@ -395,18 +395,11 @@ def generate_heatmaps(
...
-
-    if normalize:
-        # normalize all heatmaps to one
-        confidence = confidence / torch.sum(confidence, dim=(2, 3), keepdim=True)
-    else:
-        confidence /= sigma * torch.sqrt(2 * torch.tensor(np.pi))
-
+    # normalize all heatmaps to one
+    confidence = confidence / torch.sum(confidence, dim=(2, 3), keepdim=True)
+    # replace nans with zeros heatmaps
```
danbider, line 400: add a comment that "we won't try to output an all-zeros heatmap, but rather all-zeros heatmaps will be ignored in the supervised heatmap loss"

lightning_pose/data/utils.py
```diff
@@ -395,18 +395,11 @@ def generate_heatmaps(
     confidence *= -1
     confidence /= 2 * sigma**2
     confidence = torch.exp(confidence)
```
danbider, line 397: @colehurwitz what is `confidence` at this stage for a NaN keypoint? just making sure we have no surprises

danbider (changes requested):
minor suggestions and qs

lightning_pose/data/utils.py
themattinthehatt: we discussed this after you left the meeting last week - I don't see any reason why we should have this option available. it isn't used anywhere else. if we need to resurrect it at some point that's fine but for now it's just clutter.

lightning_pose/data/utils.py
themattinthehatt: i was thinking about this last time I looked at this function - we should just rename this variable `heatmap` since that's what it actually is

colehurwitz: Changed the name to heatmap. Honestly, I have no idea what confidence/heatmaps are for NaN points before zeroing.

themattinthehatt:
looks good to me!

themattinthehatt (approved):

danbider (approved):
great, happy to have this in

PR #67, opened 2022-11-13
moving metric computation into repo and pca cleanup

themattinthehatt:


lightning_pose/losses/losses.py
```diff
@@ -275,13 +273,11 @@ def __init__(
...
         super().__init__(data_module=data_module, log_weight=log_weight)
         self.loss_name = loss_name
-        self.error_metric = error_metric
 
         if loss_name == "pca_multiview":
             if mirrored_column_matches is None:
                 raise ValueError("must provide mirrored_column_matches in data config")
 
-        # TODO: solve issues with pca loss + data augmentation
         # the current data_module contains datasets that are loaded using augmentations. the
```
danbider, line 281: is this comment updated?

lightning_pose/metrics.py
```diff
@@ -0,0 +1,184 @@
...
+    """
+
+    from lightning_pose.losses.losses import TemporalLoss
+
+    t_loss = TemporalLoss()
+
+    if not isinstance(keypoints_pred, torch.Tensor):
+        keypoints_pred = torch.tensor(keypoints_pred, device=t_loss.device, dtype=torch.float32)
+
+    if len(keypoints_pred.shape) != 2:
```
danbider, line 45: can add comment 
```
# (samples, n_keypoints, 2) -> (samples, n_keypoints*2)
```

lightning_pose/metrics.py
```diff
@@ -0,0 +1,184 @@
...
+        keypoints_pred = keypoints_pred.reshape(keypoints_pred.shape[0], -1)
+
+    # compute loss with already-implemented class
+    t_norm = t_loss.compute_loss(keypoints_pred)
+    # prepend nan vector
```
danbider, line 50: can add
```
no temporal norm for the very first frame.
```

lightning_pose/metrics.py
```diff
@@ -0,0 +1,184 @@
...
+    keypoints_reproj_resize = reproj.reshape(reproj.shape[0], reproj.shape[1] // 2, 2)
+
+    # resize reprojections back to original dims
+    keypoints_reproj = _resize_keypoints(cfg, keypoints_reproj_resize, orig_to_resize=False)
+
+    # compute pixel error
+    error_pca = pixel_error(
+        keypoints_pred[:, pca_cols, :].cpu().numpy(), keypoints_reproj.cpu().numpy())
+
+    # next, put this back into a full keypoints pred arr
```
danbider, line 98:
```
# and save NaNs for keypoints not included in pose for pca
```

lightning_pose/metrics.py
```diff
@@ -0,0 +1,184 @@
...
+    error_pca = error_pca.reshape(
+        -1,
+        len(mirrored_column_matches[0]),
+        len(mirrored_column_matches),
+    )  # batch X num_used_keypoints X num_views
+    error_all = np.nan * np.zeros((original_dims[0], original_dims[1]))
+    for c, cols in enumerate(mirrored_column_matches):
+        error_all[:, cols] = error_pca[:, :, c]  # just the columns belonging to view c
+
+    return error_all
```
danbider, line 163: I don't know if a priority but the two losses are very similar at least initially. wondering if we can reuse code in some way. maybe not too urgent

lightning_pose/models/base.py
```diff
@@ -367,13 +367,13 @@ def evaluate_labeled(
         # compute and log loss on labeled data
         loss, log_list = self.loss_factory(stage=stage, **data_dict)
 
-        # compute and log rmse loss on labeled data
+        # compute and log pixel_error loss on labeled data
```
danbider, line 370: what makes the actual loss different now that we call it "pixel error"? it's the same rmse computation no?

lightning_pose/utils/pca.py
```diff
@@ -372,71 +304,6 @@ def __call__(self) -> int:
...
-
-    # print("reprojection_loss.shape: {}".format(reprojection_loss.shape))
-    # print("diff: {}".format(diff[:5, :]))
-    # print("diff_arr_per_keypoint: {}".format(diff_arr_per_keypoint[:5, :, :]))
-
-    return reprojection_loss
-
-
-@typechecked
-def add_params_to_loss_dict(
```
danbider, line 416: it's been a while since I looked at this guy. what's going on with it? do we still use it?

lightning_pose/utils/predictions.py
```diff
@@ -209,7 +209,7 @@ def add_split_indices_to_df(self, df: pd.DataFrame) -> pd.DataFrame:
         }
 
         for key, val in dataset_split_indices.items():
-            df.loc[val, "set"] = np.repeat(key, len(val))
+            df.loc[val, ("set", "", "")] = np.repeat(key, len(val))
```
danbider, line 212: better format I guess, we had issues there

scripts/train_hydra.py
```diff
@@ -172,9 +173,13 @@ def train(cfg: DictConfig):
...
     # ----------------------------------------------------------------------------------
     pretty_print_str("Predicting train/val/test images...")
+    # compute and save frame-wise predictions
+    preds_file = os.path.join(hydra_output_directory, "predictions.csv")
     predict_dataset(
         cfg=cfg, trainer=trainer, model=model, data_module=data_module_pred, ckpt_file=best_ckpt,
-        preds_file=os.path.join(hydra_output_directory, "predictions.csv"))
+        preds_file=preds_file)
+    # compute and save various metrics
+    compute_metrics(cfg=cfg, preds_file=preds_file, data_module=data_module)
```
danbider, line 182: nice!! useful!! also for vids!

tests/test_metrics.py
```diff
@@ -0,0 +1,106 @@
+"""Test metrics module."""
+
+import numpy as np
+
+
+def test_pixel_error():
```
danbider, line 6: I know `pixel_error()` is super simple but it is the same as `RegressionRMSE` no? Can we just use the latter you think?

lightning_pose/utils/scripts.py
```diff
@@ -439,6 +448,97 @@ def get_model(
...
+    cfg: DictConfig,
+    preds_file: str,
+    data_module: Optional[Union[BaseDataModule, UnlabeledDataModule]] = None,
+) -> None:
+    """Compute various metrics on predictions csv file."""
+
+    # get keypoint names
+    labels_file = os.path.join(cfg["data"]["data_dir"], cfg["data"]["csv_file"])
+    labels_df = pd.read_csv(labels_file, header=list(cfg["data"]["header_rows"]), index_col=0)
+    keypoint_names = get_keypoint_names(
```
danbider, line 462: if we input `data_module` I think we can get `keypoint_names` from `data_module.dataset.keypoint_names`
see [this line](https://github.com/danbider/lightning-pose/blob/22b6ebe69390db221743b483556242fb24acb2a1/lightning_pose/data/datasets.py#L90)

lightning_pose/utils/scripts.py
```diff
@@ -439,6 +448,97 @@ def get_model(
...
+
+    keypoints_pred = tmp[:, :, :2]  # shape (samples, n_keypoints, 2)
+    # confidences = tmp[:, :, -1]  # shape (samples, n_keypoints)
+
+    # hard-code metrics for now
+    if is_video:
+        metrics_to_compute = ["temporal"]
+    else:  # labeled data
+        metrics_to_compute = ["pixel_error"]
+    if cfg.data.get("columns_for_singleview_pca", None) is not None \
```
danbider, line 487: add comment
```
# for either labeled and unlabeled data, if one of the pca losses is specified in config, we compute the associated metric.
```

lightning_pose/utils/scripts.py
```diff
@@ -439,6 +448,97 @@ def get_model(
...
+    if "pca_singleview" in metrics_to_compute:
+        # build pca object
+        pca = KeypointPCA(
+            loss_type="pca_singleview",
+            data_module=data_module,
+            components_to_keep=cfg.losses.pca_singleview.components_to_keep,
+            empirical_epsilon_percentile=cfg.losses.pca_singleview.empirical_epsilon_percentile,
+            columns_for_singleview_pca=cfg.data.columns_for_singleview_pca,
+        )
+        pca()
```
danbider, line 518: `# re-fit pca on the labeled data to get params`

danbider (changes requested):
looks really fantastic, these all are changes we needed for a while. 
just asked for a few comments here and there.
and if you could explain to me what's different with the logged rmse/pixel error.

lightning_pose/losses/losses.py
themattinthehatt: yes, this was from before our updates a few weeks ago when there was no data aug for unlabeled data

lightning_pose/metrics.py
themattinthehatt: good idea, done

tests/test_metrics.py
themattinthehatt: let's revisit later, there's some reshaping that would need to happen that would actually obscure what is going on in this function; since we only use it for logging and not gradients i think it's fine to keep separate for now, but I understand the desire to unify

lightning_pose/utils/scripts.py
themattinthehatt: good catch! i'm gonna leave for now though since `data_module` is optional, only used for pca losses. i don't want someone to have to create a data module when computing pixel error say just to get kp names

danbider:
done, merged and branch deleted

PR #69, opened 2022-11-16
Unimodal conf filtered

danbider:
* changed configs
* made the unimodal losses handle low confidence frames
* hparam search shows no gains, but it is supported


themattinthehatt:
sooper :)

PR #76, opened 2023-02-01
changes for longer sequence context prediction

danbider:
1. longer sequence in DALI prediction, akin to base
2. differences in num_iters
3. a notebook to visualize (would be deleted later, keeping it here until everything is clear)
4. mostly identical to previous results. slight differences that aren't clear to me.
5. ibl-pupil and mirror mouse were changed 
Please verify my logic before we proceed


scripts/configs/config_ibl-pupil.yaml
```diff
@@ -107,7 +107,7 @@ model:
 
 dali:
   general:
-    num_threads: 4 # Number of threads for data loading; can be changed
+    num_threads: 1 # Number of threads for data loading; can be changed
```
themattinthehatt, line 110: should make this change for all configs

lightning_pose/utils/predictions.py
```diff
@@ -117,6 +117,14 @@ def unpack_preds(
         stacked_confs = torch.vstack([pred[1] for pred in preds])
 
         if self.video_file is not None:  # dealing with dali loaders
+            
+            # DB: this used to be an else but I think it should apply to all dataloaders now
+            # first we chop off the last few rows that are not part of the video
+            num_rows_to_discard = stacked_preds.shape[0] - self.frame_count
```
themattinthehatt, line 123: i think we might have issues when `stacked_preds.shape[0] < self.frame_count
`

themattinthehatt (changes requested):

PR #88, opened 2023-05-12
Feature/install dali

danbider:
* dali is installed upon automatic detection of cuda version (tests are passing on a cloudstudio with cuda 11.8 and nvidia-dali-cuda110 installation from pip.
* added a Dockerfile from Luca Madeiros (including some modifications). Maybe there is still redundant code there but it should run. @ankitvishnu23 please test on your end too if `pytest` goes well.


setup.py
```diff
@@ -1,12 +1,39 @@
...
+        raise FileNotFoundError("nvcc not found. Install CUDA and make sure it's in your path.")
+        return None
+
+cuda_version = get_cuda_version()
+
+if cuda_version is not None:
+    if 11.0 <= cuda_version < 12.0:
+        dali = "nvidia-dali-cuda110"
+    elif 12.0 <= cuda_version < 13.0:
+        dali = "nvidia-dali-cuda120"
```
themattinthehatt, line 30: should we fix this to be `"nvidia-dali-cuda120==1.25.0"` to match our pinned version from before?

Dockerfile
themattinthehatt, line 1: make name this `Dockerfile_cuda11` since we'll likely have a separate one for cuda 12

setup.py
danbider: for 120 it's fine. cuda110 has a different release; 1.16.0 I think, in my tests. We never had issues with DALI versions before so I don't worry too much about it... If you have strong preferences I can pin each separately

danbider:
Please see DALI issue
https://github.com/NVIDIA/DALI/issues/4849

PR #89, opened 2023-05-15
Adding Dockerfile.cuda12

ankitvishnu23:


Dockerfile.cuda12
```diff
@@ -0,0 +1,30 @@
...
+        python3-wheel \
+        python3-opencv \
+        xorg \
+        nvidia-driver-460
+
+RUN rm /usr/bin/python3 && ln -s python3.10 /usr/bin/python3
+
+RUN python3 -m pip install -U pip && python3 -m pip install -U setuptools
+
+RUN git clone https://github.com/danbider/lightning-pose
```
luca-medeiros, line 26: I don't think we need to clone if we are running from the folder that already has lightning-pose. Besides this could break versioning.

luca-medeiros (commented):
Maybe follow Dockerfile.cuda11 structure and just change the base image? If issues/improvements found on Dockerfile.cuda11 we could open a new issue

Dockerfile.cuda12
ankitvishnu23: switched to copying the folder!

ankitvishnu23: also had to make some changes because of the ubuntu version (22.04 v. 18.040 so the Dockerfile.cuda12 structure is different.

PR #106, opened 2023-08-02
Feature/vit

ankitvishnu23:


lightning_pose/models/backbones/vit_img_encoder.py
```diff
@@ -0,0 +1,124 @@
...
+            )
+
+from typing import Tuple, Type, List
+
+# This class and its supporting functions below lightly adapted from the ViTDet backbone available at: https://github.com/facebookresearch/detectron2/blob/main/detectron2/modeling/backbone/vit.py # noqa
+class ImageEncoderViT_FT(ImageEncoderViT):
+    def __init__(
+        self,
+        img_size: int = 1024,
+        finetune_img_size: int = 1024,
```
themattinthehatt, line 30: add this arg to the Args doc below

lightning_pose/models/backbones/vit_img_encoder.py
themattinthehatt, line 1: nice, this is much cleaner than before

themattinthehatt (changes requested):
changes needed:
* make sure to add unit tests back in!!
* please add to end of `install_requires` list in `setup.py`: "segment_anything @ git+https://github.com/facebookresearch/segment-anything.git",
* also add a comment above that line: # PyPI does not support direct dependencies, so we remove this line before uploading from PyPI
* make sure to run unit tests a final time after adding vit tests back in

PR #114, opened 2023-09-11
Multiview

farzad-ziaie:
I could not find the multiview branch so i sent the request to develop.


lightning_pose/data/datasets.py
```diff
@@ -330,12 +332,191 @@ def __getitem__(self, idx: int) -> HeatmapLabeledExampleDict:
...
+        self.pytorch_transform = transforms.Compose(pytorch_transform_list)
+
+        # # keypoints has been already transformed above
+
+        # Compute heatmaps as preprocessing step
+        self.num_targets = torch.numel(self.keypoints[0])
+        self.num_keypoints = self.num_targets // 2
+        self.label_heatmaps = None  # populated by `self.compute_heatmaps()`
+        self.compute_heatmaps()
+
```
themattinthehatt, line 442: add height/width property; assume it is the same for all views, so we just need to access that property from a single view

```
@property
def height(self) -> int:
    return self.datasets[self.view_names[0]].height
```

themattinthehatt:
```
@property
def num_views(self) -> int:
    return len(self.datasets)
```

themattinthehatt:
```
def __len__():
    return len(self.datasets[self.view_names[0]])
```

lightning_pose/data/utils.py
```diff
@@ -16,17 +16,25 @@ class BaseLabeledExampleDict(TypedDict):
...
+    keypoints: TensorType["num_targets", float]
     idxs: int
 
 
 class HeatmapLabeledExampleDict(BaseLabeledExampleDict):
     """Return type when calling __getitem__() on HeatmapTrackingDataset."""
 
-    heatmaps: TensorType["num_keypoints", "heatmap_height", "heatmap_width"]
+    heatmaps: TensorType["num_keypoints", "heatmap_height", "heatmap_width", float]
+
```
themattinthehatt, line 30: add `MultiviewLabeledExampleDict` - for now this should look exactly like `BaseLabeledExampleDict`

lightning_pose/data/utils.py
```diff
@@ -16,17 +16,25 @@ class BaseLabeledExampleDict(TypedDict):
...
+class MultiviewHeatmapLabeledExampleDict(BaseLabeledExampleDict):
```
themattinthehatt, line 31: make this inherit from `MultiviewLabeledExampleDict`, and add heatmaps (just copying `HeatmapLabeledExampleDict` for now). We'll add some more info to this later down the line.

lightning_pose/utils/scripts.py
```diff
@@ -84,6 +85,15 @@ def get_dataset(
             do_context=cfg.model.model_type == "heatmap_mhcrnn" or cfg.model.do_context,
             uniform_heatmaps=cfg.training.get("uniform_heatmaps_for_nan_keypoints", False),
         )
+    elif cfg.model.model_type == "MultiviewHeatmapDataset":
```
themattinthehatt, line 88: should move this inside the above `elif` statement, so that it looks like this:
```
elif cfg.model.model_type == "heatmap" or cfg.model.model_type == "heatmap_mhcrnn":
    if cfg.data.get(“view_names”, None) and len(cfg.data.view_names) > 1:
        dataset = MultiviewHeatmapDataset(...)
    else:
        dataset = HeatmapDataset(...)
```

This allows the `cfg.model.model_type` field to continue representing the model type irrespective of whether or not we do multiview; we might want a simple multiview heatmap model, or a more complex multiview heatmap_mhcrnn model.

themattinthehatt: for now we are not supporting multiview operations with the regression model, so we should also do:
```
if cfg.model.model_type == "regression":
    if cfg.data.get(“view_names”, None) and len(cfg.data.view_names) > 1:
        raise NotImplementedError("Multi-view support only available for heatmap-based models")
    else:
        dataset = BaseTrackingDataset(...)
```

lightning_pose/utils/scripts.py
```diff
@@ -161,7 +171,7 @@ def get_loss_factories(
 
     # collect all supervised losses in a dict; no extra params needed
     # set "log_weight = 0.0" so that weight = 1 and effective weight is (1 / 2)
-    if cfg.model.model_type == "heatmap" or cfg.model.model_type == "heatmap_mhcrnn":
+    if cfg.model.model_type == "heatmap" or cfg.model.model_type == "heatmap_mhcrnn" or cfg.model.model_type == "MultiviewHeatmapDataset":
```
themattinthehatt, line 174: can remove the extra `or cfg.model.modely_type == "MutliviewHeatmapDataset"` now

lightning_pose/utils/scripts.py
```diff
@@ -254,7 +264,7 @@ def get_model(
                 do_context=cfg.model.do_context,
                 image_size=image_h,  # only used by ViT
             )
-        elif cfg.model.model_type == "heatmap":
+        elif cfg.model.model_type == "heatmap" or cfg.model.model_type == "MultiviewHeatmapDataset":
```
themattinthehatt, line 267: can remove extra `or` statement now

scripts/configs/config_7m.yaml
```diff
@@ -1,234 +1,141 @@
...
+    width: 256
+    height: 256
+  # ABSOLUTE path to data directory. the below is just a toy example for running Getting Started scripts
+  data_dir: /mnt/scratch2/farzad/7m/data
+  # ABSOLUTE path to unlabeled videos' directory. the below is just a toy example for running Getting Started scripts
+  video_dir: videos
   # location of labels; for example script, this should be relative to `data_dir`
-  csv_file: master_multicam_labeledFrames.csv
-  # downsample heatmaps - 2 | 3
+  csv_file:
```
themattinthehatt, line 15: @danbider what do you think about enumerating all of the actual csv file names vs programmatically assigning them based on a root csv file name and the view names? forces more structure on the data (which people may not like) but also reduces room for error.

scripts/configs/config_7m.yaml
themattinthehatt, line 1: i definitely want to include this yaml file in the repo as a multi-view example, but I'll ask that you store the actual yaml file you use for experiments somewhere outside of the github repo. that way as you run experiments and fiddle around with the yaml file those changes don't get constantly pushed to the repo.

scripts/converters/c7mtolp.py
themattinthehatt, line 1: for now let's keep this code outside of the repo since it pertains to a single dataset. maybe later we'll add converters for DPK and SLEAP, but in general I want to keep the repo as lean as possible (and therefore easier to maintain).

themattinthehatt (changes requested):
great start @farzad-ziaie! hopefully the suggestions I made make sense and will further simplify the new code. 

also, before the next PR, check out the contributing guidelines here:
https://github.com/danbider/lightning-pose/blob/main/docs/contributing.md

running flake8, isort, and pytest before each PR will ensure code style and content stay in good shape along the way.

lightning_pose/data/datasets.py
```diff
@@ -330,12 +332,191 @@ def __getitem__(self, idx: int) -> HeatmapLabeledExampleDict:
...
+    def compute_heatmaps(self):
```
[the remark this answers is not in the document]
farzad-ziaie: the main heatmap class uses super().__getitem__(idx) method which is not usable for multiview heatmap. I understand the logic behind it, i think you wanted to apply polymorphism here but for that we need to refactor the heatmapdataset as well.

lightning_pose/data/datasets.py
```diff
@@ -330,12 +332,191 @@ def __getitem__(self, idx: int) -> HeatmapLabeledExampleDict:
...
+        if len(self.imgaug_transform) == 1 and isinstance(self.imgaug_transform[0], iaa.Resize):
             # we have a deterministic resizing augmentation; use precomputed heatmaps
             example_dict["heatmaps"] = self.label_heatmaps[idx]
         else:
             # we have a random augmentation; need to recompute heatmaps
-            example_dict["heatmaps"] = self.compute_heatmap(example_dict)
+            example_dict["heatmaps"] = self.compute_heatmap(example_dict)   
         return example_dict
+
+# class MultiviewHeatmapDataset(torch.utils.data.Dataset):
```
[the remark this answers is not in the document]
farzad-ziaie: i am using some of the methods from HeatmapDataset which are exactly the same as MultiviewHeatmapDataset so i tought it would be ok to inheret from heatmap, should i replicate the methods and inheret from torch.utils.data.Dataset?

lightning_pose/data/datasets.py
```diff
@@ -330,12 +332,191 @@ def __getitem__(self, idx: int) -> HeatmapLabeledExampleDict:
...
+                should be in the form
+                (image_path, bodypart_1_x, bodypart_1_y, ..., bodypart_n_y)
+                Note: image_path is relative to the given root_directory
+            header_rows: which rows in the csv are header rows
+            imgaug_transform: imgaug transform pipeline to apply to images
+            downsample_factor: factor by which to downsample original image dims to have a smaller
+                heatmap
+            do_context: include additional frames of context if possible
+        """
+        self.root_directory = root_directory
```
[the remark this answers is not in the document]
farzad-ziaie: great idea! composition instead of inheretance.

lightning_pose/data/datasets.py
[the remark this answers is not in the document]
themattinthehatt: can you clarify - why isn't this usable for multiview heatmap? we just want to loop over single-view heatmap datasets and have each load their own view data. in this process, the single-view HeatmapDatasets will each create their own heatmaps and return them to the MultiviewHeatmapDataset. So the MultiviewHeatmapDataset shouldn't need to worry about computing heatmaps itself.

lightning_pose/data/datasets.py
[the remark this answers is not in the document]
themattinthehatt: @farzad-ziaie did you have another question on this one? I saw a comment in the PR email that isn't showing up on github

lightning_pose/data/datasets.py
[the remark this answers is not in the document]
farzad-ziaie: I replied to this before reading the rest, removed this part and followed your instruction after it.

PR #115, opened 2023-09-25
Multi view

farzad-ziaie:


lightning_pose/data/datasets.py
```diff
@@ -335,3 +339,184 @@ def __getitem__(self, idx: int) -> HeatmapLabeledExampleDict:
...
+        """
+        heatmaps = []
+        keypoints = []
+        images = []
+        concat_order = []
+        for view_num, (view, data) in enumerate(datadict.items()):
+            heatmap_key_num = data["heatmaps"].shape[0]
+            heatmap_length = data["heatmaps"].shape[1]
+            heatmap_width = data["heatmaps"].shape[2]
+            heatmap_buffer = torch.zeros(heatmap_key_num,
```
themattinthehatt, line 484: this part can be simplified - basically, we just want to concatenate along the keypoint dimension. so if view 0 has 4 keypoints and view 1 has 8 keypoints, then the final heatmap tensor just needs to be of shape (12, heatmap_height, heatmap_width) where 12 = 4 + 8; same with keypoints.

so you should be able to do something like
```
heatmaps = torch.cat([d["heatmaps"] for _, d in datadict.items()])
```

for the images, even if they start with a different initial size we are going to resize them all the same. so the final image tensor will be of shape (num_views, 3, image_height, image_width)

lightning_pose/data/utils.py
```diff
@@ -25,8 +25,27 @@ class BaseLabeledExampleDict(TypedDict):
...
     """Return type when calling __getitem__() on HeatmapTrackingDataset."""
+    heatmaps: TensorType["num_keypoints", "heatmap_height", "heatmap_width", float]
+
+
+class MultiviewLabeledExampleDict(TypedDict):
+    """Return type when calling __getitem__() on BaseTrackingDataset."""
+    images: Union[
+        TensorType["RGB":3, "image_height", "image_width", float],
+        TensorType["frames", "RGB":3, "image_height", "image_width", float],
+    ]
```
themattinthehatt, line 36: each will have "num_views" as the first entry, followed by what currently exists

lightning_pose/data/utils.py
```diff
@@ -25,8 +25,27 @@ class BaseLabeledExampleDict(TypedDict):
...
+
+
+class MultiviewHeatmapLabeledExampleDict(MultiviewLabeledExampleDict):
+    """Return type when calling __getitem__() on MultiviewHeatmapDataset."""
 
     heatmaps: TensorType["num_keypoints", "heatmap_height", "heatmap_width", float]
+    # bbx: TensorType["bounding_box", float]
+    num_views: int
+    concat_order: List[int]
+    view_names: List[int]
```
themattinthehatt, line 50: need to add `MultiviewLabeledBatchDict` and `MultiviewHeatmapLabeledBatchDict` - note these are BATCH dicts instead of EXAMPLE dicts

lightning_pose/models/base.py
```diff
@@ -16,6 +16,7 @@
     SemiSupervisedBatchDict,
     SemiSupervisedHeatmapBatchDict,
     UnlabeledBatchDict,
+    MultiviewHeatmapLabeledExampleDict,
```
themattinthehatt, line 19: BatchDict instead of ExampleDict

lightning_pose/models/base.py
```diff
@@ -269,14 +270,16 @@ class BaseSupervisedTracker(BaseFeatureExtractor):
 
     def get_loss_inputs_labeled(
         self,
-        batch_dict: Union[BaseLabeledBatchDict, HeatmapLabeledBatchDict],
+        batch_dict: Union[BaseLabeledBatchDict, HeatmapLabeledBatchDict,
+                          MultiviewHeatmapLabeledExampleDict],
```
themattinthehatt, line 274: BatchDict instead of ExampleDict

lightning_pose/models/base.py
```diff
@@ -269,14 +270,16 @@ class BaseSupervisedTracker(BaseFeatureExtractor):
...
     ) -> dict:
         """Return predicted coordinates for a batch of data."""
         raise NotImplementedError
 
     def evaluate_labeled(
         self,
-        batch_dict: Union[BaseLabeledBatchDict, HeatmapLabeledBatchDict],
+        batch_dict: Union[BaseLabeledBatchDict, HeatmapLabeledBatchDict,
+                          MultiviewHeatmapLabeledExampleDict],
```
themattinthehatt, line 282: BatchDict instead of ExampleDict

lightning_pose/models/base.py
```diff
@@ -303,7 +306,8 @@ def evaluate_labeled(
 
     def training_step(
         self,
-        batch_dict: Union[BaseLabeledBatchDict, HeatmapLabeledBatchDict],
+        batch_dict: Union[BaseLabeledBatchDict, HeatmapLabeledBatchDict,
+                          MultiviewHeatmapLabeledExampleDict],
```
themattinthehatt, line 310: BatchDict instead of ExampleDict

lightning_pose/models/base.py
```diff
@@ -312,15 +316,17 @@ def training_step(
 
     def validation_step(
         self,
-        batch_dict: Union[BaseLabeledBatchDict, HeatmapLabeledBatchDict],
+        batch_dict: Union[BaseLabeledBatchDict, HeatmapLabeledBatchDict,
+                          MultiviewHeatmapLabeledExampleDict],
```
themattinthehatt, line 320: BatchDict instead of ExampleDict

lightning_pose/models/base.py
```diff
@@ -312,15 +316,17 @@ def training_step(
...
         batch_idx: int,
     ) -> None:
         """Base validation step, a wrapper around the `evaluate_labeled` method."""
         self.evaluate_labeled(batch_dict, "val")
 
     def test_step(
         self,
-        batch_dict: Union[BaseLabeledBatchDict, HeatmapLabeledBatchDict],
+        batch_dict: Union[BaseLabeledBatchDict, HeatmapLabeledBatchDict,
+                          MultiviewHeatmapLabeledExampleDict],
```
themattinthehatt, line 329: BatchDict instead of ExampleDict

lightning_pose/utils/predictions.py
```diff
@@ -304,7 +323,12 @@ def predict_dataset(
 
     pred_handler = PredictionHandler(cfg=cfg, data_module=data_module, video_file=None)
     labeled_preds_df = pred_handler(preds=labeled_preds)
-    labeled_preds_df.to_csv(preds_file)
+    if isinstance(labeled_preds_df, dict):
+        for view_name, df in labeled_preds_df.items():
+            view_name = "_" + view_name + ".csv"
+            df.to_csv(preds_file.replace(".csv", view_name))
```
themattinthehatt, line 329: instead of view_name = "_" + ...

can do

df.to_csv(preds_file.replace(".csv", f"_{view_name}.csv"))

scripts/train_hydra.py
```diff
@@ -33,6 +33,10 @@ def train(cfg: DictConfig):
     print("Our Hydra config file:")
     pretty_print_cfg(cfg)
 
+    if cfg.model.model_type == "regression":
```
themattinthehatt, line 36: can remove this check, already taken care of in dataset creation

themattinthehatt (changes requested):

lightning_pose/utils/scripts.py
```diff
@@ -250,16 +270,22 @@ def get_model(
                 image_size=image_h,  # only used by ViT
             )
         elif cfg.model.model_type == "heatmap":
+            if cfg.model.get("backbone_pretrained", None):
```
themattinthehatt, line 273: can shorten this to
```
backbone_pretrained = cfg.model.get("backbone_pretrained", True)
```

tests/conftest.py
```diff
@@ -91,6 +219,25 @@ def heatmap_dataset(cfg, imgaug_transform) -> HeatmapDataset:
     torch.cuda.empty_cache()
 
 
+@pytest.fixture
+def MultiviewHeatmap_Dataset(cfg_multiview, imgaug_transform) -> MultiviewHeatmapDataset:
```
themattinthehatt, line 223: rename function/fixture to `multiview_heatmap_dataset`

tests/conftest.py
```diff
@@ -152,6 +299,27 @@ def heatmap_data_module(cfg, heatmap_dataset) -> BaseDataModule:
     torch.cuda.empty_cache()
 
 
+@pytest.fixture
+def Multiview_heatmap_data_module(cfg, MultiviewHeatmap_Dataset) -> BaseDataModule:
```
themattinthehatt, line 303: keep naming convention; change to `multiview_heatmap_data_module`

tests/data/test_datamodules.py
```diff
@@ -60,6 +60,32 @@ def test_heatmap_datamodule(cfg, heatmap_data_module):
     del batch
 
 
+def test_Multiview_heatmap_datamodule(cfg_multiview, Multiview_heatmap_data_module):
```
themattinthehatt, line 63: same, lowercase "m"

tests/data/test_datasets.py
```diff
@@ -36,8 +36,27 @@ def test_heatmap_dataset(cfg, heatmap_dataset):
     assert batch["images"].shape == (3, im_height, im_width)
     assert batch["keypoints"].shape == (num_targets,)
     assert batch["heatmaps"].shape[1:] == heatmap_dataset.output_shape
-    assert type(batch["images"]) == torch.Tensor
-    assert type(batch["keypoints"]) == torch.Tensor
+    assert type(batch["images"]) is torch.Tensor
+    assert type(batch["keypoints"]) is torch.Tensor
+
+
+def test_Multiview_heatmap_dataset(cfg_multiview, MultiviewHeatmap_Dataset):
```
themattinthehatt, line 43: same, lowercase "m"

lightning_pose/models/backbones/torchvision.py
```diff
@@ -112,6 +113,10 @@ def build_backbone(
         # load resnet or efficientnet models from torchvision.models
         base = getattr(tvmodels, backbone_arch)(weights=weights)
 
+        if not pretrained and backbone_arch == "resnet50":
```
themattinthehatt, line 116: i thought we agreed to not proceed this way - we want to keep the backbone intact, and not change the number of channels in the first layer. instead, take a batch of shape (batch, n_view, 3, height, width), reshape to (batch * n_views, 3, height, width), push through the network, then reshape back to (bach, n_view, n_keypoints, height, width)

themattinthehatt (changes requested):
@farzad-ziaie  getting closer - still one descrepancy between what's implemented here and what we discussed.

we want to keep the backbone intact, and not change the number of channels in the first layer. instead, we want 

1. the multiview dataset to create a batch of shape (batch, num_views, 3, height, width)
2. have the model's `forward` function check to see if the batch contains multiple views or not (by checking `len(images.shape)`)
3. if multiple views, reshape to (batch * num_views, 3, height, width); if not multiple views don't need to do anything
4. push (possibly reshaped) batch through the network by running `get_representations` and `heatmaps_from_representations` (i.e. leave as-is)
5. if multiview, at this point you have a tensor of shape (bach * num_views, num_keypoints, height, width). this needs to be reshaped into a tensor of shape (batch, num_views * num_keypoints, height, width) - from that point forward all of the unsupervised loss computation will (should) work out-of-the-box. 

if any part of that is not clear let me know and we can meet to discuss further!

lightning_pose/data/datasets.py
```diff
@@ -335,3 +340,175 @@ def __getitem__(self, idx: int) -> HeatmapLabeledExampleDict:
...
+        return (
+            self.height // 2**self.downsample_factor,
+            self.width // 2**self.downsample_factor,
+        )
+
+    @property
+    def num_views(self) -> int:
+        return len(self.view_names)
+
+    def fusion(self, datadict: dict, bbox: Optional[np.array] = np.array((0))):
```
themattinthehatt, line 468: add return type:
```
def fusion(
    self, 
    datadict: dict, 
    bbox: Optional[np.array] = np.array((0))
) -> Tuple[
    Union[
        TensorType["num_views", "RGB":3, "image_height", "image_width", float],
        TensorType["num_views", "frames", "RGB":3, "image_height", "image_width", float],
    ],    
    (add others here)
]:
```
(will also need to add `from typing import Tuple` to top)

lightning_pose/data/utils.py
```diff
@@ -25,8 +25,28 @@ class BaseLabeledExampleDict(TypedDict):
 
 class HeatmapLabeledExampleDict(BaseLabeledExampleDict):
     """Return type when calling __getitem__() on HeatmapTrackingDataset."""
+    heatmaps: TensorType["num_keypoints", "heatmap_height", "heatmap_width", float]
+
+
+class MultiviewLabeledBatchDict(TypedDict):
```
themattinthehatt, line 31: should be `MultiviewLabeledExampleDict` (rather than *Batch*Dict)

lightning_pose/data/utils.py
```diff
@@ -25,8 +25,28 @@ class BaseLabeledExampleDict(TypedDict):
...
+    """Return type when calling __getitem__() on BaseTrackingDataset."""
+    images: Union[
+        TensorType["num_views", "RGB":3, "image_height", "image_width", float],
+        TensorType["num_views", "frames", "RGB":3, "image_height", "image_width", float],
+    ]
+    keypoints: TensorType["num_targets", float]
+    idxs: int
+
+
+class MultiviewHeatmapLabeledBatchDict(MultiviewLabeledBatchDict):
```
themattinthehatt, line 41: change *Batch* to *Example* in both class name and inheritance

lightning_pose/data/datasets.py
```diff
@@ -335,3 +340,175 @@ def __getitem__(self, idx: int) -> HeatmapLabeledExampleDict:
...
+        Images and their heatmaps and then stacks them.
+        """
+        datadict = {}
+        img_sizes = {}
+        for view in self.view_names:
+            datadict[view] = self.dataset[view][idx]
+            img_sizes[view] = self.dataset[view].image_original_size[idx]
+        images, heatmaps, keypoints, concat_order = self.fusion(datadict)
+
+        return MultiviewHeatmapLabeledBatchDict(
```
themattinthehatt, line 506: Batch -> Example (see below)

lightning_pose/data/utils.py
themattinthehatt: will then need to add, below `HeatmapLabeledBatchDict`,
```
MultiviewLabeledBatchDict
```
and
```
MultiviewHeatmapLabeledBatchDict
```
which will mostly look like the above, but with additional "batch" dimensions on the tensors

lightning_pose/models/heatmap_tracker.py
```diff
@@ -260,12 +264,21 @@ def heatmaps_from_representations(
...
-        images: TensorType["batch", "frames", "channels":3, "image_height", "image_width"],
+        images: Union[TensorType["batch", "channels":3, "image_height", "image_width"],
+                      TensorType["batch", "views", "channels":3, "image_height", "image_width"]]
     ) -> TensorType["num_valid_outputs", "num_keypoints", "heatmap_height", "heatmap_width"]:
         """Forward pass through the network."""
         # we get one representation for each desired output.
-        representations = self.get_representations(images)
-        heatmaps = self.heatmaps_from_representations(representations)
+        shape=images.shape
+        if len(shape)>4:
```
themattinthehatt, line 273: add a comment here just saying that if len(shape) > 4 we assume we have multiple views and need to combine images across batch/views before passing to network, then we reshape

themattinthehatt (changes requested):
getting close!!

lightning_pose/data/datasets.py
farzad-ziaie: should i also change dataset.py? line 501?

themattinthehatt (approved):

Issue #86, opened 2023-05-03
dependency on fixed image size


thejanzimmermann:
Hi there, amazing work. I am wondering how vital the dependency on fixed image sizes for training your setup is. Reading the code it seems you are currently requiring a fixed image size but DALIs transform module should be able to scale dynamically.

Best

Jan

themattinthehatt:
hey @thejanzimmermann, thanks for checking out the repo! Currently we do require fixed image sizes; this was a design choice we made early on because our datasets did not contain images of different sizes. It's not conceptually crucial that image sizes are fixed, but it would require a bit of work to generalize this in the repo. Are you working with multiple image sizes in a single dataset?

thejanzimmermann:
Thanks so much. For training yes. Our use cases are usually cropped images and since our arenas are large a fixed size makes it difficult.
________________________________

themattinthehatt:
@thejanzimmermann are you interested in trying lightning pose out on your data? We don't have the bandwidth at the moment to support this feature (though we'd like to in the future), but if you'd like to try it yourself we'd be happy to meet and talk about what would be necessary for a PR.

danbider:
@thejanzimmermann how does your data look like? labeled images of different shapes? or unlabeled videos of different shapes? or both? we're well aware of your work and it'll be cool to collaborate with you here a bit.

in principle, this requirement could be dropped -- we do all of the meaningful computations on the resized image shapes (and resizing happens on the fly). But fixing it needs the careful attention that we don't currently have...

videos are assumed to have the same shape so that we could undo the data augmentation and compute our unsupervised losses in the unified resized image shape (orig image dims parameter could be dropped too if we dynamically keep track of video size in DALI).

I think that a PR here will need to track the original image shapes both in 1) imgaug labeled frame loading , 2) DALI unlabeled video loading, and in `model.predict_step()`.

another simple option (just to know if lightning-pose is worth your time): you can preprocess the images and labels; resize the images and keypoint labels to a standard size, and train.

themattinthehatt:
closed by #117

PR #117, opened 2023-10-20
update code to use original image and video pixel coordinates for all predictions

Selmaan:
this applies to supervised & unsupervised training, and inference. Image size is also no longer part of a configuration. Image sizes are always computed on the fly, and passed as an entry in batch_dicts (bbox). This is used to transform predictions from resized to original coords. This should make the code long-term compatible with upcoming changes for a 2-stage dynamic-crop algorithm. At present, it only affects the code by rescaling all losses computed per-pixel, e.g. a temporal loss threshold of 20 pixels no applies to 20 pixels in the original, not resized, coords.

### DIFF
--- a/lightning_pose/data/augmentations.py
+++ b/lightning_pose/data/augmentations.py
@@ -97,11 +97,12 @@ def imgaug_transform(cfg: DictConfig) -> iaa.Sequential:
     else:
         raise NotImplementedError("must choose imgaug kind from 'default', 'dlc', 'dlc-top-down'")
 
-    # always apply resizing transform
-    data_transform.append(
-        iaa.Resize({
-            "height": cfg.data.image_resize_dims.height,
-            "width": cfg.data.image_resize_dims.width}
-        ))
+    # do not resize when using dynamic crop pipeline
+    if not cfg.data.get('dynamic_crop', False):
+        data_transform.append(
+            iaa.Resize({
+                "height": cfg.data.image_resize_dims.height,
+                "width": cfg.data.image_resize_dims.width}
+            ))
 
     return iaa.Sequential(data_transform)
--- a/lightning_pose/data/dali.py
+++ b/lightning_pose/data/dali.py
@@ -158,9 +158,9 @@ class LitDaliWrapper(DALIGenericIterator):
         # shape (1,) or (2, 3)
         transforms = batch[0]["transforms"][0]
         # get frame size, order is batch,seq_len,H,W,C
-        height = batch[0]["frame_size"][0,1]
-        width = batch[0]["frame_size"][0,2]
-        bbox = torch.tensor([0,0,height,width]).repeat((frames.shape[0],1))
+        height = batch[0]["frame_size"][0, 1]
+        width = batch[0]["frame_size"][0, 2]
+        bbox = torch.tensor([0, 0, height, width]).repeat((frames.shape[0], 1))
 
         return UnlabeledBatchDict(frames=frames, transforms=transforms, bbox=bbox)
 
--- a/lightning_pose/data/dali.py
+++ b/lightning_pose/data/dali.py
@@ -63,6 +63,7 @@ def video_pipe(
     Returns:
         pipeline object to be fed to DALIGenericIterator
         data augmentation matrix (returned so that geometric transforms can be undone)
+        size of video frames, used for bbox
 
     """
     video = fn.readers.video(
@@ -80,6 +81,7 @@ def video_pipe(
         pad_last_batch=pad_last_batch,  # Important for context loaders
         file_list_include_preceding_frame=True,  # to get rid of dali warnings
     )
+    orig_size = fn.shapes(video)
     if resize_dims:
         video = fn.resize(video, size=resize_dims)
     if imgaug == "dlc" or imgaug == "dlc-top-down":
@@ -115,7 +117,7 @@ def video_pipe(
         mean=normalization_mean,
         std=normalization_std,
     )
-    return transform, matrix
+    return transform, matrix, orig_size
 
 
 class LitDaliWrapper(DALIGenericIterator):
@@ -155,8 +157,12 @@ class LitDaliWrapper(DALIGenericIterator):
         frames = batch[0]["frames"][0, :, :, :, :]
         # shape (1,) or (2, 3)
         transforms = batch[0]["transforms"][0]
+        # get frame size, order is batch,seq_len,H,W,C
+        height = batch[0]["frame_size"][0,1]
+        width = batch[0]["frame_size"][0,2]
+        bbox = torch.tensor([0,0,height,width]).repeat((frames.shape[0],1))
 
-        return UnlabeledBatchDict(frames=frames, transforms=transforms)
+        return UnlabeledBatchDict(frames=frames, transforms=transforms, bbox=bbox)
 
     def __next__(self) -> UnlabeledBatchDict:
         batch = super().__next__()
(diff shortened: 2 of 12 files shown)


lightning_pose/data/datasets.py
```diff
@@ -260,8 +261,6 @@ def __init__(
         # Compute heatmaps as preprocessing step
         self.num_targets = torch.numel(self.keypoints[0])
         self.num_keypoints = self.num_targets // 2
-        self.label_heatmaps = None  # populated by `self.compute_heatmaps()`
-        self.compute_heatmaps()
```
themattinthehatt, line 264: nice

lightning_pose/data/datasets.py
```diff
@@ -328,10 +327,156 @@ def __getitem__(self, idx: int) -> HeatmapLabeledExampleDict:
 
         """
         example_dict: BaseLabeledExampleDict = super().__getitem__(idx)
-        if len(self.imgaug_transform) == 1 and isinstance(self.imgaug_transform[0], iaa.Resize):
-            # we have a deterministic resizing augmentation; use precomputed heatmaps
-            example_dict["heatmaps"] = self.label_heatmaps[idx]
```
themattinthehatt, line 333: thanks for ditching this, should have happened long ago

lightning_pose/data/datasets.py
```diff
@@ -328,10 +327,156 @@ def __getitem__(self, idx: int) -> HeatmapLabeledExampleDict:
...
+        example_dict["heatmaps"] = self.compute_heatmap(example_dict)
+        return example_dict
+
+
+class DetectorDataset(BaseTrackingDataset):
```
themattinthehatt, line 334: cleaner if you inherit from `HeatmapDataset` - then you don't need to to redefine `compute_heatmap`, `output_shape`, etc., as well as some of the attribute setting in `__init__` (`output_sigma`, `uniform_heatmaps`, etc.).

you can still overwrite `__getitem__` to do whatever you want

lightning_pose/models/heatmap_tracker.py
```diff
@@ -269,16 +269,40 @@ def forward(
...
+            predicted_keypoints[:, :, 1] /= batch_dict["images"].shape[-2]  # -2 dim is height "y"
+        else:  # we have unlabeled dict, 'frames' instead of 'images'
+            predicted_keypoints[:, :, 0] /= batch_dict["frames"].shape[-1]  # -1 dim is width "x"
+            predicted_keypoints[:, :, 1] /= batch_dict["frames"].shape[-2]  # -2 dim is height "y"
+        # multiply and add by bbox dims (x,y,h,w)
+        for i in range(predicted_keypoints.shape[0]):
+            predicted_keypoints[i, :, 0] *= batch_dict["bbox"][i, 3]  # scale x by box width
+            predicted_keypoints[i, :, 0] += batch_dict["bbox"][i, 0]  # add bbox x offset
+            predicted_keypoints[i, :, 1] *= batch_dict["bbox"][i, 2]  # scale y by box height
+            predicted_keypoints[i, :, 1] += batch_dict["bbox"][i, 1]  # add bbox y offset
```
themattinthehatt, line 289: can this be vectorized? should check if something like `predicted_keypoints[:, :, 0] *= batch_dict["bbox"][:, 3]` works (might need `batch_dict["bbox"][:, 3].unsqueeze(1)` or something)

scripts/configs/config_birdCOM.yaml
```diff
@@ -0,0 +1,213 @@
...
+  # max allowed number of instances, i.e. 1 for single-animal, >1 for multiple animal
+  num_max_instances: 1
+  # detector network params
+
+training:
+  # select from one of several predefined image/video augmentation pipelines
+  # default- resizing only
+  # dlc- imgaug pipeline implemented in DLC 2.0 package
+  # dlc-top-down- dlc augmentations plus vertical and horizontal flips
+  imgaug: dlc-top-down
```
themattinthehatt, line 32: fyi for some reason i found `dlc-top-down` to be worse than `dlc` for a top-down two-mouse dataset - later on i'd be interested in how these compare with your dataset

themattinthehatt (changes requested):
@Selmaan PR looks great! I added some comments, take a look and let me know if you have any questions. besides the comments it would be great if you could add a unit test for `convert_bbox_coords` - then I think we can wrap this up

lightning_pose/data/augmentations.py
```diff
@@ -97,11 +97,12 @@ def imgaug_transform(cfg: DictConfig) -> iaa.Sequential:
     else:
         raise NotImplementedError("must choose imgaug kind from 'default', 'dlc', 'dlc-top-down'")
 
-    # always apply resizing transform
-    data_transform.append(
-        iaa.Resize({
-            "height": cfg.data.image_resize_dims.height,
-            "width": cfg.data.image_resize_dims.width}
-        ))
+    # do not resize when using dynamic crop pipeline
```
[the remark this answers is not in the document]
Selmaan: This is a little tricky to handle, because we haven't totally fleshed out the way the pipeline is going to work? But I think we are planning to have just one dataset, in which case we will want access to the full, un-resized images, although I'm not sure yet how to do this in an efficient way keeping the images on the GPU

lightning_pose/data/dali.py
```diff
@@ -155,8 +157,12 @@ def _dali_output_to_tensors(batch: list) -> UnlabeledBatchDict:
         frames = batch[0]["frames"][0, :, :, :, :]
         # shape (1,) or (2, 3)
         transforms = batch[0]["transforms"][0]
+        # get frame size, order is batch,seq_len,H,W,C
```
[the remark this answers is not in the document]
Selmaan: thx, yea I copied that from DALI documentation but I think you're right, will change

scripts/configs/config_birdCOM.yaml
Selmaan: that is weird and something I never compared! In my old code I had a L-R flip but not a top-down flip. Interested to see what matters

scripts/configs/config_birdCOM_backup.yaml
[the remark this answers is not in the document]
Selmaan: I'll delete this in a future commit

lightning_pose/models/heatmap_tracker.py
```diff
@@ -269,16 +269,40 @@ def forward(
         # softmax temp stays 1 here; to modify for model predictions, see constructor
         return spatial_softmax2d(heatmaps, temperature=torch.tensor([1.0]))
 
+    def convert_bbox_coords(self, batch_dict: Union[HeatmapLabeledBatchDict, UnlabeledBatchDict],
```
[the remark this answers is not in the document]
Selmaan: will make this change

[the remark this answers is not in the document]
themattinthehatt: sounds good - can also pass those in as args

the main thing is i want a unit test for this function, since it's an important one

lightning_pose/models/heatmap_tracker.py
Selmaan: unsqueeze works!

lightning_pose/models/heatmap_tracker.py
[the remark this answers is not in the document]
Selmaan: will do that next. all other changes should be fixed in the new commit below

themattinthehatt:
closes #86

PR #119, opened 2023-11-17
[add] compute_metrics for multiview, it is a loop for each file plus file name check

farzad-ziaie:


themattinthehatt (approved):
very nice 👌

PR #157, opened 2024-05-23
ignore hidden "_labeled" folders in dlc to lp conversion

hummuscience:
The conversion script was failing as it was detecting folders in the "labeled_frames" folder that did not contain any frame data.

Examples of such folders are hidden folders, such as .DS_Store folders that MacOS generates.
The other examples are folders created by DeepLabCut when running the `check_labels` function (folders where labels are plotted on top of the frames).

This is a simple fix to filter out the potential problematic folders.

There is maybe a better/safer approach by generating a list of videos (from the videos folder or the config.yaml file) and then looking for folders in the labeled_videos folder that are exact matches of the basenames.

### DIFF
--- a/scripts/converters/dlc2lp.py
+++ b/scripts/converters/dlc2lp.py
@@ -24,7 +24,9 @@ if dlc_dir == lp_dir:
     raise NameError(f"dlc_dir and lp_dir cannot be the same")
 
 # find all labeled data in DLC project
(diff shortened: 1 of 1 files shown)


themattinthehatt:
@hummuscience thanks so much for this PR! I'll take a look at this tomorrow and get back to you :pray:

PR #180, opened 2024-06-27
Passing vfr_check to fn.readers.video

hummuscience:
Here is the corrected pull request with the variable passed to the correct function. 

(continued from here: https://github.com/danbider/lightning-pose/pull/179)


themattinthehatt (approved):
thanks!

Issue #138, opened 2024-04-07
how to carry out multi-GPU with lightning-pose?


Wulin-Tan:
Hi, lighting pose team
I am very interested in this super cool tool.
How to do multi-GPU with lighting pose?
I can't find multi-GPU information in the tutorial.

themattinthehatt:
We do not currently support multi-GPU training but it should not be difficult to implement with our current setup (at least for the supervised models). We'll be happy to look into this but it might take us a week or so, we'll keep you updated.

YitingChang:
I'm also interested in multi-GPU training and would like to follow this. Thank you!

YitingChang:
Hi @ksikka , 

Thank you for following up! I'm encountering memory limitations with a single GPU, so I'd like to explore running Lightning Pose with multiple GPUs.

To give you more context on my current setup: as I add more features—such as unsupervised losses and temporal context networks—the memory demands increase. I've had to significantly resize images and reduce batch sizes to fit within the memory constraints. Currently, I’m running Lightning Pose on a cluster with one GPU that has 40 GB of memory. Allocating more GPUs shouldn't be an issue if Lightning Pose can support multi-GPU functionality.

Wulin-Tan:
@ksikka yes, I need to handle more GPU memory as well as faster training if possible.

ksikka:
Hi @YitingChang and @Wulin-Tan support for multi-GPU was just added for supervised training. This works for both heatmap and heatmap_mchrnn as long as `losses_to_use: []`. Support for multi-gpu when using unsupervised losses is in the works.

ksikka:
Without multi-gpu on unsupervised, you might have success with gradient accumulation (see `accumulate_grad_batches` in the [configuration file docs](https://lightning-pose.readthedocs.io/en/latest/source/user_guide/config_file.html). Say you want an effective batch size of 8 but can only fit batch size 2 in memory. then you could set `train_batch_size: 2` and `accumulate_grad_batches: 4`. This is an experimental feature so if you do try it, report back with your feedback. Thanks!

ksikka:
Hi @YitingChang and @Wulin-Tan support for multi-GPU was just added for unsupervised training and is now ready for you to try out! We've tested it and been pleased with the results. Please see the guide here, and report back if it has helped you.
https://lightning-pose.readthedocs.io/en/latest/source/user_guide_advanced/multi_gpu_training.html

YitingChang:
[...]

PR #205, opened 2024-10-15
Change backbone finetuning strategy to allow for DDP

ksikka:
See lightning-ai/pytorch-lightning#20340 for the original issue. The implemented workaround no longer freezes/unfreezes layers via require_grad, instead we only modulate the learning rate of the backbone layer.

Unblocks #138

### DIFF
--- a/lightning_pose/callbacks.py
+++ b/lightning_pose/callbacks.py
@@ -42,3 +42,73 @@ class AnnealWeight(Callback):
             # Dan: removed buffer; seems to complicate checkpoint loading
             # pl_module.register_buffer(self.attr_name, torch.tensor(value))
             setattr(pl_module, self.attr_name, torch.tensor(value))
+
+
+class UnfreezeBackbone(Callback):
+    """Callback that ramps up the backbone learning rate from 0 to `upsampling_lr` on
+    `unfreeze_epoch`.
+
+    Starts LR at `initial_ratio * upsampling_lr`. Grows lr by a factor of `epoch_ratio` per
+    epoch. Once LR reaches `upsampling_lr`, keeps it in sync with `upsampling_lr`.
+
+    Use instead of pl.callbacks.BackboneFinetuning in order to use multi-GPU (DDP). See
+    lightning-ai/pytorch-lightning#20340 for context.
+    """
+
+    def __init__(
+        self,
+        unfreeze_epoch,
+        initial_ratio=0.1,
+        epoch_ratio=1.5,
+    ):
+        self.unfreeze_epoch = unfreeze_epoch
+        self.initial_ratio = initial_ratio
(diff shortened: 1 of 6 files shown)


ksikka:
image

Run with and without code changes:
Capture

lightning_pose/callbacks.py
```diff
@@ -42,3 +42,61 @@ def on_train_epoch_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningMo
...
+        assert optimizer.param_groups[1]["name"].startswith("upsampling")
+
+        backbone_lr = optimizer.param_groups[0]["lr"]
+        upsampling_lr = optimizer.param_groups[1]["lr"]
+
+        optimizer.param_groups[0]["lr"] = self._get_next_backbone_lr(
+            pl_module.current_epoch, backbone_lr, upsampling_lr
+        )
+
+    def _get_next_backbone_lr(self, current_epoch, backbone_lr, upsampling_lr):
```
danbider, line 86: good to comment on the `if` conditions, what is this the LR schedule that is implemented here

lightning_pose/models/base.py
```diff
@@ -589,39 +580,3 @@ def training_step(
         self.log("total_loss", total_loss, prog_bar=True)
 
         return {"loss": total_loss}
-
```
danbider, line 592: good cleanup

tests/test_callbacks.py
```diff
@@ -0,0 +1,22 @@
...
+
+def test_unfreeze_backbone():
+    unfreeze_backbone = UnfreezeBackbone(2)
+
+    # Test unfreezing at epoch 2.
+    assert unfreeze_backbone._get_next_backbone_lr(0, 0.0, 1e-3) == 0.0
+    assert unfreeze_backbone._get_next_backbone_lr(1, 0.0, 1e-3) == 0.0
+    assert unfreeze_backbone._get_next_backbone_lr(2, 0.0, 1e-3) == 1e-4
+
+    # Test warming up.
```
danbider, line 11: comment on the 1.5 factor?

tests/test_callbacks.py
```diff
@@ -0,0 +1,22 @@
+from lightning_pose.callbacks import UnfreezeBackbone
+
+def test_unfreeze_backbone():
+    unfreeze_backbone = UnfreezeBackbone(2)
+
+    # Test unfreezing at epoch 2.
```
danbider, line 6: comment that expected behavior is zero lr until epoch 2? then specify why 1e-4. which is the set LR?

danbider (approved):
very happy to see these updates -- thanks @ksikka!

themattinthehatt (approved):

PR #211, opened 2024-11-01
Added option to choose colormap for labelled videos

hummuscience:
I wanted to change the standard colormap for the labelled videos ("cool") is it is harder to differentiate keypoints when the number is larger. So I added the option to choose the colormap in the yaml file.

### DIFF
--- a/lightning_pose/utils/scripts.py
+++ b/lightning_pose/utils/scripts.py
@@ -756,4 +756,5 @@ def export_predictions_and_labeled_video(
             ys_arr=ys_arr,
             mask_array=mask_array,
             filename=labeled_mp4_file,
+            colormap=cfg.eval.colormap
         )
--- a/lightning_pose/utils/scripts.py
+++ b/lightning_pose/utils/scripts.py
@@ -756,5 +756,5 @@ def export_predictions_and_labeled_video(
             ys_arr=ys_arr,
             mask_array=mask_array,
             filename=labeled_mp4_file,
-            colormap=cfg.eval.colormap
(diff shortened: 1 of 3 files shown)


lightning_pose/utils/scripts.py
```diff
@@ -756,4 +756,5 @@ def export_predictions_and_labeled_video(
             ys_arr=ys_arr,
             mask_array=mask_array,
             filename=labeled_mp4_file,
+            colormap=cfg.eval.colormap
```
themattinthehatt, line 759: update to `colormap=cfg.eval.get("colormap", "cool")` in case someone runs this with an older config that doesn't have the `colormap` option

themattinthehatt (changes requested):

lightning_pose/utils/scripts.py
hummuscience: Good point. I was wondering how one resolves this issue. Will upodate it :)

hummuscience:
It might make sense to update [the config file docs](https://lightning-pose.readthedocs.io/en/latest/source/user_guide/config_file.html) at some point with all the other options (including the colormap).

I always liked[ this page from suite2p](https://suite2p.readthedocs.io/en/latest/settings.html) that listed all possible settings so things like `training.uniform_heatmaps_for_nan_keypoints`
are easier to find

themattinthehatt:
@hummuscience thanks for the PR - would you mind submitting a new issue about the missing config options? It's a good idea to update the docs, I can take care of that soon.

PR #214, opened 2024-11-06
Add PCA Centering method (experimental)

ksikka:
Config:

losses.pca_singleview.centering_method = null | mean | median

### DIFF
--- a/lightning_pose/losses/losses.py
+++ b/lightning_pose/losses/losses.py
@@ -284,6 +284,7 @@ class PCALoss(Loss):
         data_module: Optional[Union[BaseDataModule, UnlabeledDataModule]] = None,
         log_weight: float = 0.0,
         device: Union[Literal["cuda", "cpu"], torch.device] = _DEFAULT_TORCH_DEVICE,
+        centering_method: Optional[Literal["mean", "median"]] = None,
         **kwargs,
     ) -> None:
         super().__init__(data_module=data_module, log_weight=log_weight)
@@ -310,6 +311,7 @@ class PCALoss(Loss):
             mirrored_column_matches=mirrored_column_matches,
             columns_for_singleview_pca=columns_for_singleview_pca,
             device=device,
+            centering_method=centering_method,
         )
         # compute all the parameters needed for the loss
         self.pca()
--- a/lightning_pose/utils/pca.py
+++ b/lightning_pose/utils/pca.py
@@ -38,6 +38,7 @@ class KeypointPCA(object):
         mirrored_column_matches: Optional[Union[ListConfig, List]] = None,
         columns_for_singleview_pca: Optional[Union[ListConfig, List]] = None,
         device: Union[Literal["cuda", "cpu"], torch.device] = "cpu",
+        centering_method: Optional[Literal["mean", "median"]] = None,
     ) -> None:
         self.loss_type = loss_type
         self.data_module = data_module
@@ -61,6 +62,7 @@ class KeypointPCA(object):
         self.columns_for_singleview_pca = columns_for_singleview_pca
         self.pca_object = None
         self.device = device
+        self.centering_method = centering_method
 
     def _get_data(self) -> None:
         self.data_arr, _ = DataExtractor(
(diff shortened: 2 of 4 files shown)


themattinthehatt (approved):
looks good!

hummuscience:
Quick question, does centering in this case solve the problem of the animal having different "sizes" when images come from different datasets?

ksikka:
Hi Muad, this is different. It was an attempt to improve the PCA loss that did not produce the expected results, so it remains experimental.

Aside from this PR, we're actively workshopping a two-stage pipeline that detects and crops the animal, then passes it to the pose estimation model. However in our current plans, the crop size was fixed. It's interesting to learn about the use-case of animals being different sizes from different datasets. 

To handle animals of different sizes, perhaps to some extent this could be handled by image augmentation? I.e. randomly cropping the image prior to the resize transform (which resizes to a fixed height and width). This would effectively scale the animal. 
https://github.com/paninski-lab/lightning-pose/blob/main/lightning_pose/data/augmentations.py#L119

hummuscience:
I thought that LP already implemented scaling augmentation through dlc_top_down, but that does not seem to be the case. I was surprised because all the other augmentations are implemented. Until I realized that DLC does "scale augmentation" not through imgaug but through resizing all images to a sampled resizing factor for each batch: https://github.com/DeepLabCut/DeepLabCut/blob/main/deeplabcut/pose_estimation_tensorflow/datasets/pose_imgaug.py

I wonder what would be the best way to do that. The nice thing about the DLC approach is that the random rescaling happens for every batch.

hummuscience:
I think I will give this a try in the pipeline:

```
data_transform.append(
            iaa.Sometimes(
                1,  # probability of applying scaling
                iaa.Affine(
                    scale=(0.5, 1.5),
                )
            )
)
```

This should resize each batch between 50 and 150%. Since the images will be resized back to a fixed height/width, it will probably mainly be affected by the 0.4 probability of cropping.

What do you think?

themattinthehatt:
I'll note that we implement a version of scaling using `CropAndPad` in the current imgaug pipeline, which can scale the image larger or smaller: https://github.com/paninski-lab/lightning-pose/blob/2b99f81ca17b510f36d7915805efe3ffcec979d8/lightning_pose/data/augmentations.py#L113

hummuscience:
If I understand the LP dlc-top-down augmentation correctly, in 40% of the cases, the image is cropped by up to 15% from each side and due to the keep_size=False, it leads to a resizing of the image due to the final transformation, right?

The DLC augmentation does an additional step in every batch and that is the general scaling, set with scale_jitter_up/down (https://github.com/DeepLabCut/DeepLabCut/blob/d4da23a44c4e969eac437c1425b7670bc6c4ce14/deeplabcut/pose_estimation_tensorflow/datasets/pose_imgaug.py#L50).

It is a general "augmentation" that is done before any of the augmentation options where it samples from the scaling range (https://github.com/DeepLabCut/DeepLabCut/blob/d850b5e70c6c1905d5f53bf0069ac0eda680e37d/deeplabcut/pose_estimation_tensorflow/datasets/pose_base.py#L33) and then applies that to the data before any augmentation (https://github.com/DeepLabCut/DeepLabCut/blob/d4da23a44c4e969eac437c1425b7670bc6c4ce14/deeplabcut/pose_estimation_tensorflow/datasets/pose_imgaug.py#L304). 

I am only discussing this as I am trying to figure out why I am getting these "side switches" in the predictions. This happens only in models which I train using the backbone I trained from the Superanimal dataset from DLC (https://github.com/paninski-lab/lightning-pose/issues/158#issuecomment-2447795436). This dataset has the animals in very different "size" and I am wondering if the difference in scaling between DLC and LP is what is causing it.

themattinthehatt:
Ah I see. One question I have is whether or not you've looked at the outputs of the DLC Superanimal model - does it have the same side-switching problem? If not then that would suggest the issue lies in a difference between DLC and LP; if so that would suggest another potential issue, either with both models or with the dataset itself.

themattinthehatt:
Interesting - quite unexpected, I would say. Are you interested in adding a scaling augmentation to the pipeline to see if that resolves the problem?

hummuscience:
I am currently testing a scaling augmentation (dlc-top-down vs. dlc-top-down-scale) and will report.

PR #216, opened 2024-11-07
Context frames can use any delimiter

ksikka:
As an added bonus this should now work on windows paths. (prior, we were splitting on "/")

### DIFF
--- a/lightning_pose/data/datasets.py
+++ b/lightning_pose/data/datasets.py
@@ -493,7 +493,7 @@ class MultiviewHeatmapDataset(torch.utils.data.Dataset):
         )
 
 
-def _get_context_img_paths(center_img_path):
+def _get_context_img_paths(center_img_path: Path) -> list[Path]:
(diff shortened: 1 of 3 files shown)


lightning_pose/data/datasets.py
```diff
@@ -525,3 +491,35 @@ def __getitem__(self, idx: int) -> MultiviewHeatmapLabeledExampleDict:
             concat_order=concat_order,  # List[str]
             view_names=self.view_names,  # List[str]
         )
+
+
+def _get_context_img_paths(center_img_path):
```
themattinthehatt, line 496: add typehint for `center_img_path`

themattinthehatt (changes requested):
looks great! just one request for a type hint

lightning_pose/data/datasets.py
ksikka: Added

themattinthehatt (approved):

PR #221, opened 2024-11-16
Fix tests and post-training analysis behavior

ksikka:


themattinthehatt:
@ksikka can you describe what it is you are fixing here?

scripts/predict_new_vids.py
```diff
@@ -61,7 +61,7 @@ def predict_videos_in_dir(cfg: DictConfig):
         absolute_cfg_path = return_absolute_path(hydra_relative_path, n_dirs_back=2)
 
         # load model
-        model_cfg = OmegaConf.load(os.path.join(absolute_cfg_path, ".hydra/config.yaml"))
+        model_cfg = OmegaConf.load(os.path.join(absolute_cfg_path, "config.yaml"))
```
ksikka, line 64: Another fix here, the original code loads the config without CLI overrides - hydra stores overrides separately in .hydra/overrides.yaml

themattinthehatt: good catch!

lightning_pose/train.py
```diff
@@ -187,6 +187,13 @@ def train(cfg: DictConfig) -> None:
     else:
         data_module_pred = data_module
 
+    model = load_model_from_checkpoint(
```
themattinthehatt, line 190: is the goal here to ensure that the "model" object floating around has the correct weights loaded, rather than relying on proper use of "best_ckpt" below? If so you should also remove "best_ckpt" arg from L250 `export_predictions_and_labeled_video` (since otherwise we'll waste a bit of time loading the best weights into the model that already has the best weights loaded)

themattinthehatt (changes requested):

themattinthehatt (approved):

PR #222, opened 2024-11-18
Save copy of labeled data csv to output dir

ksikka:
Gives memory of what labeled datapoints were used to train the model, for use in future features.

Additional changes
* Save files before training starts. This way they exist even if training prematurely dies.
* Use resolve=True flag of OmegaConf.save which will resolve ${vars} to their strings in the saved config.

### DIFF
--- a/lightning_pose/train.py
+++ b/lightning_pose/train.py
@@ -2,8 +2,10 @@
 
 import os
 import random
+import shutil
 import sys
 import warnings
+from pathlib import Path
 
 import lightning.pytorch as pl
 import numpy as np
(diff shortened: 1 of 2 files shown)


ksikka:
Depends on the "Fix is_file" commit from #221

lightning_pose/train.py
```diff
@@ -181,17 +204,14 @@ def train(cfg: DictConfig) -> None:
         preds_file=preds_file,
     )
     # compute and save various metrics
-    try:
-        # take care of multiview case, where multiple csv files have been saved
-        preds_files = [
-            os.path.join(hydra_output_directory, path) for path in
-            os.listdir(hydra_output_directory) if path.endswith(".csv")
```
ksikka, line 188: Reason for changing this was that path.endswith(".csv") was picking up the saved labeled csv files.

themattinthehatt (approved):

PR #230, opened 2024-12-19
get rid of unnecessary lightning logs during inference

themattinthehatt:
just a minor cleanup, tired of randomly finding folders with hundreds of log files scattered about my computer :)

### DIFF
--- a/lightning_pose/utils/predictions.py
+++ b/lightning_pose/utils/predictions.py
@@ -329,7 +329,7 @@ def predict_dataset(
 
     delete_trainer = False
     if trainer is None:
-        trainer = pl.Trainer(devices=1, accelerator="auto")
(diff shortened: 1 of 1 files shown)


ksikka:
Nice! With this change we should also remove passing `trainer` to `predict_dataset` and `export_predictions_and_labeled_video` in `train.py`, so that the prediction function is responsible for creating the trainer and your new logic will trigger.

Further we can mark the trainer kwargs as deprecated or remove it altogether depending on how much of a breaking change you think this will be for clients.

ksikka:
PR is also fine as is, if you just want to merge and close for now.

themattinthehatt:
I'm fine leaving trainer kwarg in there for now, I would rather not introduce any breaking changes at this point - but we can definitely add this to the list of LP 2.0 updates

PR #232, opened 2025-01-07
Split up train, add Model class, add litpose CLI

ksikka:
New model class will be the new programmatic interface to lightning pose model prediction. utils/predictions.py can be made private in 2.0.

To test `litpose` locally, `git pull` then `pip install -e ".[dev]"`.

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -155,11 +155,7 @@ def _predict(args: argparse.Namespace):
     # Delay this import because it's slow.
     from lightning_pose.model import Model
 
-    model_dir = Path(args.model_dir)
-    if not model_dir.is_dir():
-        raise FileNotFoundError(f"Model directory not found: {model_dir.absolute()}")
-
-    model = Model.from_dir(model_dir)
+    model = Model.from_dir(args.model_dir)
     input_paths = [Path(p) for p in args.input_path]
 
     for p in input_paths:
--- a/cli/main.py
+++ b/cli/main.py
@@ -1,0 +1,197 @@
+from __future__ import annotations
+
+import argparse
+import datetime
+import os
+import sys
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from lightning_pose.model import Model
+
+from . import friendly, types
+
+
+def _build_parser():
+    parser = friendly.ArgumentParser()
+    subparsers = parser.add_subparsers(
+        dest="command",
+        required=True,
+        help="Litpose command to run.",
+        parser_class=friendly.ArgumentSubParser,
+    )
+
+    # Train command
+    train_parser = subparsers.add_parser(
+        "train",
+        description="Train a lightning-pose model using the specified configuration file.",
+        usage="litpose train <config_file> \\\n"
+        "                      [--output_dir OUTPUT_DIR] \\\n"
+        "                      [--overrides KEY=VALUE...]"
+        "",
+    )
+    train_parser.add_argument(
+        "config_file",
+        type=types.config_file,
+        help="path a config file.\n"
+        "Download and modify the config template from: \n"
+        "https://github.com/paninski-lab/lightning-pose/blob/main/scripts/configs/config_default.yaml",
+    )
+    train_parser.add_argument(
+        "--output_dir",
+        type=types.model_dir,
+        help="explicitly specifies the output model directory.\n"
+        "If not specified, defaults to "
+        "./outputs/{YYYY-MM-DD}/{HH:MM:SS}/",
+    )
+    train_parser.add_argument(
+        "--overrides",
+        nargs="*",
+        metavar="KEY=VALUE",
+        help="overrides attributes of the config file. Uses hydra syntax:\n"
+        "https://hydra.cc/docs/advanced/override_grammar/basic/",
+    )
+
+    # Add arguments specific to the 'train' command here
+
+    # Predict command
+    predict_parser = subparsers.add_parser(
+        "predict",
+        description="Predicts keypoints on videos or images.\n"
+        "\n"
+        "  Video predictions are saved to:\n"
+        "    <model_dir>/\n"
+        "    └── video_preds/\n"
+        "        ├── <video_filename>.csv              (predictions)\n"
+        "        ├── <video_filename>_<metric>.csv     (losses)\n"
+        "        └── labeled_videos/\n"
+        "            └── <video_filename>_labeled.mp4\n"
+        "\n"
+        "  Image predictions are saved to:\n"
+        "    <model_dir>/\n"
+        "    └── image_preds/\n"
+        "        └── <image_dirname | csv_filename | timestamp>/\n"
+        "            ├── predictions.csv\n"
+        "            ├── predictions_<metric>.csv      (losses)\n"
+        "            └── <image_filename>_labeled.png\n",
+        usage="litpose predict <model_dir> <input_path:video|image|dir|csv>...  [OPTIONS]",
+    )
+    predict_parser.add_argument(
+        "model_dir", type=types.existing_model_dir, help="path to a model directory"
+    )
+
+    predict_parser.add_argument(
+        "input_path",
+        type=Path,
+        nargs="+",
+        help="one or more paths. They can be video files, image files, CSV files, or directories.\n"
+        "    directory: predicts over videos or images in the directory.\n"
+        "               saves image outputs to `image_preds/<directory_name>`\n"
+        "    video file: predicts on the video\n"
+        "    image file: predicts on the image. saves outputs to `image_preds/<timestamp>`\n"
+        "    CSV file: predicts on the images specified in the file.\n"
+        "              uses the labels to compute pixel error.\n"
+        "              saves outputs to `image_preds/<csv_file_name>`\n",
+    )
+
+    post_prediction_args = predict_parser.add_argument_group("post-prediction")
+    post_prediction_args.add_argument(
+        "--skip_viz",
+        action="store_true",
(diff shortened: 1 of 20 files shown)


lightning_pose/train.py
```diff
@@ -37,13 +26,112 @@
...
+            model.config.cfg.data.csv_file, model.config.cfg.data.view_names
+        ):
+            csv_files.append(
+                _absolute_csv_file(csv_file, model.config.cfg.data.data_dir)
+            )
+            output_filename_stems.append(f"predictions_{view_name}")
+
+    for csv_file, output_filename_stem in zip(csv_files, output_filename_stems):
+        model.predict_on_label_csv_internal(
+            csv_file=csv_file,
```
ksikka, line 76: This doesn't work yet for multiview. I tried to pass a single csv_file here, but in the multiview case, something dowstream tries to create a multiview dataset which requires multiple csv files.

lightning_pose/utils/predictions.py
```diff
@@ -95,14 +93,7 @@ def frame_count(self) -> int:
...
     def keypoint_names(self):
-        if self.cfg.data.get("keypoint_names", None) is not None:
-            if isinstance(self.cfg.data.get("keypoint_names"), DictConfig):
-                return dict(self.cfg.data.get("keypoint_names"))
-            return list(self.cfg.data.keypoint_names)
-        elif self.cfg.data.get("keypoints", None) is not None:
-            return list(self.cfg.data.keypoints)
-        else:
-            return self.data_module.dataset.keypoint_names
+        return list(self.cfg.data.keypoint_names)
```
ksikka, line 96: Here and above: PR-unrelated simplification

lightning_pose/utils/scripts.py
```diff
@@ -553,21 +554,29 @@ def compute_metrics(
...
-        preds_file = sorted(preds_file)
         for view_name, csv_file, preds_file_ in zip(
                 sorted(cfg.data.view_names),
                 sorted(cfg.data.csv_file),
-                preds_file
+                sorted(preds_file)
         ):
             assert view_name in preds_file_
             labels_file = return_absolute_path(os.path.join(cfg.data.data_dir, csv_file))
-            # preds_file_ = preds_file.replace(".csv", f"_{view_name}.csv")
```
ksikka, line 570: Here and above: PR-unrelated simplification

lightning_pose/utils/scripts.py
```diff
@@ -651,7 +661,7 @@ def compute_metrics_single(
         # add train/val/test split
         if set is not None:
             error_df["set"] = set
-        save_file = preds_file.replace(".csv", "_pixel_error.csv")
+        save_file = preds_file_path.with_name(preds_file_path.stem + "_pixel_error.csv")
```
ksikka, line 664: This is because when you call `litpose predict <model_dir> <csv_file>` it outputs in `image_preds/CSV_FILE_NAME/*` and we need to avoid replacing the `.csv` in `CSV_FILE_NAME`

lightning_pose/train.py
```diff
@@ -37,13 +26,112 @@
...
+    return csv_file
+
+
+def _evaluate_on_training_dataset(model: Model):
+    pretty_print_str("Predicting train/val/test images...")
+
+    if model.config.is_single_view():
+        csv_file = _absolute_csv_file(
+            model.config.cfg.data.csv_file, model.config.cfg.data.data_dir
+        )
```
ksikka, line 60: _absolute_csv_file is a bunch of logic to handle the (possibly imaginary?) case that csv_file is an absolute path instead of relative to data_dir.

lightning_pose/model.py
```diff
@@ -0,0 +1,255 @@
...
+                compute_metrics_on_preds_file = []
+                for view_name in sorted(self.config.cfg.data.view_names):
+                    multiview_preds_file_path = preds_file_path.with_name(
+                        preds_file_path.stem + f"_{view_name}.csv"
+                    )
+                    compute_metrics_on_preds_file.append(str(multiview_preds_file_path))
+
+            import os
+
+            print(os.getcwd())
```
ksikka, line 182: Remove

cli/main.py
```diff
@@ -0,0 +1,197 @@
...
+        # Maintain legacy hydra chdir until downstream no longer depends on it.
+        os.chdir(output_dir)
+        train(cfg)
+
+
+def _predict(args: argparse.Namespace):
+    # Delay this import because it's slow.
+    from lightning_pose.model import Model
+
+    model_dir = Path(args.model_dir)
```
ksikka, line 158: This and following exists check shouldn't be necessary. It's using existing_model_dir from types

lightning_pose/train.py
ksikka: Fixed, see the commit log in this PR

cli/main.py
```diff
@@ -0,0 +1,193 @@
...
+    ):
+        cfg = hydra.compose(config_name=args.config_file.stem, overrides=args.overrides)
+
+        # Delay this import because it's slow.
+        from lightning_pose.train import train
+
+        # TODO: Move some aspects of directory mgmt to the train function.
+        output_dir.mkdir(parents=True, exist_ok=True)
+        # Maintain legacy hydra chdir until downstream no longer depends on it.
+        os.chdir(output_dir)
```
themattinthehatt, line 150: do we need to change back to the original directory after training is complete?

lightning_pose/model.py
```diff
@@ -0,0 +1,251 @@
...
+    def image_preds_dir(self):
+        return self.model_dir / "image_preds"
+
+    def video_preds_dir(self):
+        return self.model_dir / "video_preds"
+
+    def labeled_videos_dir(self):
+        return self.model_dir / "video_preds" / "labeled_videos"
+
+    UNSPECIFIED = "unspecified"
```
themattinthehatt, line 70: add comment on use of this; move it above `from_dir` method?

lightning_pose/model_config.py
themattinthehatt, line 1: this is great. at some point you should start migrating the different cfg checks in here as well; then the constructor could look like:
```
    def __init__(self, cfg: DictConfig):
        self.cfg = cfg
        self._check_cfg()

```

lightning_pose/train.py
themattinthehatt: I think it's good to leave this in place, not too much overhead for the added flexibility

lightning_pose/train.py
```diff
@@ -169,175 +258,6 @@ def train(cfg: DictConfig) -> None:
...
-    else:
-        data_module_pred = data_module
-
-    model = load_model_from_checkpoint(
-        cfg=cfg,
-        ckpt_file=best_ckpt,
-        eval=True,
-        data_module=data_module_pred,
-    )
+    from lightning_pose.model import Model
```
themattinthehatt, line 261: unnecessary import?

themattinthehatt (changes requested):
great updates :ok_hand:

lightning_pose/model_config.py
ksikka: Neat idea! Instead of in the constructor, I'd have a validate method. Occasionally you might have use for an invalid ModelConfig object.

cli/main.py
ksikka: Good question.

Originally interpreted your question as, do we need to chdir after training + post-training prediction is complete. The answer is no, as long as the command ends here and there's no subsequent logic. The process will just exit.

Can we chdir back to the original working directory once training is complete, but before the post-prediction starts? I think we could, but I haven't tested this. Does all the downstream prediction code work if the current working directory is not the model directory? 

Can we remove the chdir altogether and still train the model to the desired directory? Perhaps, if we configure logging and ModelCheckpointing correctly, and add a output_dir kwarg to train.

It's something to investigate further at some point, but for now I'll leave it as is, and add tests for prediction where current working dir != model dir, and model_dir is a relative path.

lightning_pose/train.py
ksikka: Ack.

lightning_pose/model.py
ksikka: done

lightning_pose/train.py
ksikka: done

cli/main.py
ksikka: > Does all the downstream prediction code work if the current working directory is not the model directory?

Yes, confirmed this by running `litpose predict` from some random locations and it all works as expected.

ksikka: Here's an interesting question.

Say you instantiate `Model.from_dir("some/relative/path")`

Then you os.chdir to some new location.

Should the model be relative to the original location at time of instantiation, or to the current location? 

I think probably the former. In order to do that, we need to resolve the relative path to an absolute one in the constructor of Model. I'll do that now.

themattinthehatt: yes, I agree with you - should immediately resolve the relative path to absolute in the constructor

themattinthehatt (approved):

lightning_pose/model.py
```diff
@@ -0,0 +1,254 @@
...
+        df = export_predictions_and_labeled_video(
+            video_file=str(video_file),
+            cfg=self.config.cfg,
+            prediction_csv_file=str(prediction_csv_file),
+            labeled_mp4_file=labeled_mp4_file,
+            model=self.model,
+        )
+
+        # FIXME: This is only used for computing PCA metrics.
+        data_module = _build_datamodule_pred(self.cfg)
```
themattinthehatt, line 233: should move this inside the if statement so it's not constructed unnecessarily

ksikka: Fixed

PR #234, opened 2025-01-16
Fix issues reading labels file with empty first row

### LINKED ISSUE #210: pandas incorrectly reads labels file when first row is empty
Hi, I faced the following error when trying to run multi-stream models: 

using dlc image augmentation pipeline
Error executing job with overrides: []
Traceback (most recent call last):
  File "/home/gangliaguardian/lightning-pose/scripts/train_hydra.py", line 35, in train_model
    train(cfg)
  File "/home/gangliaguardian/lightning-pose/lightning_pose/train.py", line 54, in train
    dataset = get_dataset(cfg=cfg, data_dir=data_dir, imgaug_transform=imgaug_transform)
  File "/home/gangliaguardian/lightning-pose/lightning_pose/utils/scripts.py", line 103, in get_dataset
    dataset = MultiviewHeatmapDataset(
  File "/home/gangliaguardian/lightning-pose/lightning_pose/data/datasets.py", line 382, in __init__
    self.dataset[view] = HeatmapDataset(
  File "/home/gangliaguardian/lightning-pose/lightning_pose/data/datasets.py", line 245, in __init__
    super().__init__(
  File "/home/gangliaguardian/lightning-pose/lightning_pose/data/datasets.py", line 85, in __init__
    self.keypoints = self.keypoints.reshape(self.keypoints.shape[0], -1, 2)
RuntimeError: shape '[19, -1, 2]' is invalid for input of size 361 

Here's the Config File: 
callbacks:
  anneal_weight:
    attr_name: total_unsupervised_importance
    init_val: 0.0
    increase_factor: 0.01
    final_val: 1.0
    freeze_until_epoch: 0

dali:
  general:
    seed: 123456

  base:
    train:
      sequence_length: 32
    predict:
      sequence_length: 96

  context:
    train:
      batch_size: 16
    predict:
      sequence_length: 96

data:
  # resize dimensions to streamline model creation
  # resize dimensions: we do this internally to accelerate training. make sure it's a multiple of 128.
  image_orig_dims:
    height: 540
    width: 720
  image_resize_dims:
    height: 384
    width: 384
  # ABSOLUTE path to data directory
  data_dir: /home/gangliaguardian/lightning-pose/data/test_chocolate_07_24/
  # ABSOLUTE path to unlabeled videos' directory
  video_dir: /home/gangliaguardian/lightning-pose/data/test_chocolate_07_24/videos/
  csv_file: 
    - view0.csv
    - view1.csv
    - view2.csv
  view_names:
    - view0
    - view1
    - view2
  # downsample heatmaps - 2 | 3
  downsample_factor: 2
  # total number of keypoints
  num_keypoints: 15
  # indicator to use 2-stage dynamic crop algorithm
  dynamic_crop: False
  # max allowed number of instances, i.e. 1 for single-animal, >1 for multiple animal
  num_max_instances: 1
  # keypoint names
  keypoint_names:
  - thumb_left
  - thumb_right
  - finger_1_left
  - finger_1_right
  - finger_2_left
  - finger_2_right
  - finger_3_left
  - finger_3_right
  - finger_4_left
  - finger_4_right
  - snout
  - mouth
  - dorsal_left
  - dorsal_right
  - head
  mirrored_column_matches: null

training:
  imgaug: dlc
  train_batch_size: 8
  val_batch_size: 1
  test_batch_size: 1
  train_prob: 0.95
  val_prob: 0.05
  train_frames: 1
  num_gpus: 1
  log_every_n_steps: 150
  check_val_every_n_epoch: 150
  num_workers: 4
  early_stop_patience: 3
  unfreezing_epoch: 20
  min_epochs: 500
  max_epochs: 500
  ckpt_every_n_epochs: 100
  gpu_id: 0
  rng_seed_data_pt: 0
  rng_seed_model_pt: 0
  lr_scheduler: multisteplr
  lr_scheduler_params:
    multisteplr:
      milestones: [150, 200, 250]
      gamma: 0.5

model:
  losses_to_use: []
  backbone: resnet50_animal_ap10k
  model_type: heatmap
  heatmap_loss_type: mse
  model_name: test
  checkpoint: null

losses:
  pca_multiview:
    log_weight: 5.0
    components_to_keep: 3
    epsilon: null
  pca_singleview:
    log_weight: 5.0
    components_to_keep: 0.99
    epsilon: null
  temporal:
    log_weight: 5.0
    epsilon: 20.0
    prob_threshold: 0.05

eval:
  hydra_paths: [" "]
  predict_vids_after_training: true
  save_vids_after_training: true
  fiftyone:
    dataset_name: test
    model_display_names: ["test_model"]
    launch_app_from_script: false
    remote: true
    address: 127.0.0.1
    port: 5151
  test_videos_directory: /home/gangliaguardian/lightning-pose/data/test_chocolate_07_24/videos/
  confidence_thresh_for_vid: 0.90

hydra:
  run:
    dir: data/test_1_final/models/${now:%Y-%m-%d}/${now:%H-%M-%S}
  sweep:
    dir: data/test/test_1_final/multirun/${now:%Y-%m-%d}/${now:%H-%M-%S}
    subdir: ${hydra.job.num}

ksikka:
Fixes #210

Root cause pandas-dev/pandas#21995

### DIFF
--- a/lightning_pose/data/datasets.py
+++ b/lightning_pose/data/datasets.py
@@ -2,7 +2,6 @@
 
 
 import os
-import re
 from pathlib import Path
 from typing import Callable, List, Literal, Optional, Tuple, Union
 
@@ -20,6 +19,7 @@ from lightning_pose.data.utils import (
     MultiviewHeatmapLabeledExampleDict,
     generate_heatmaps,
 )
+from lightning_pose.utils import io as io_utils
 from lightning_pose.utils.io import get_context_img_paths, get_keypoint_names
 
 # to ignore imports for sphix-autoapidoc
@@ -77,6 +77,7 @@ class BaseTrackingDataset(torch.utils.data.Dataset):
             raise FileNotFoundError(f"Could not find csv file at {csv_file}!")
 
         csv_data = pd.read_csv(csv_file, header=header_rows, index_col=0)
+        csv_data = io_utils.fix_empty_first_row(csv_data)
         self.keypoint_names = get_keypoint_names(csv_file=csv_file, header_rows=header_rows)
         self.image_names = list(csv_data.index)
         self.keypoints = torch.tensor(csv_data.to_numpy(), dtype=torch.float32)
@@ -492,4 +493,3 @@ class MultiviewHeatmapDataset(torch.utils.data.Dataset):
             concat_order=concat_order,  # List[str]
             view_names=self.view_names,  # List[str]
         )
-
--- a/lightning_pose/utils/fiftyone.py
+++ b/lightning_pose/utils/fiftyone.py
@@ -9,6 +9,7 @@ from PIL import Image
 from tqdm import tqdm
 from typeguard import typechecked
 
+from lightning_pose.utils import io as io_utils
 from lightning_pose.utils import pretty_print_str
 from lightning_pose.utils.io import return_absolute_data_paths, return_absolute_path
 
@@ -81,13 +82,16 @@ class FiftyOneImagePlotter:
             df_tmp = []
             csv_files = [os.path.join(self.data_dir, f) for f in self.cfg.data.csv_file]
             for csv_file in csv_files:
-                df_tmp.append(pd.read_csv(csv_file, header=self.df_header_rows))
+                csv_data = pd.read_csv(csv_file, header=self.df_header_rows)
+                csv_data = io_utils.fix_empty_first_row(csv_data)
+                df_tmp.append(csv_data)
             self.ground_truth_df = pd.concat(df_tmp)
         else:
             self.ground_truth_df: pd.DataFrame = pd.read_csv(
                 os.path.join(self.data_dir, self.cfg.data.csv_file),
                 header=self.df_header_rows,
             )
+            self.ground_truth_df = io_utils.fix_empty_first_row(self.ground_truth_df)
         if self.keypoints_to_plot is None:
             # plot all keypoints that appear in the ground-truth dataframe
             self.keypoints_to_plot: List[str] = list(self.ground_truth_df.columns.levels[0])
--- a/lightning_pose/utils/scripts.py
+++ b/lightning_pose/utils/scripts.py
@@ -39,6 +39,7 @@ from lightning_pose.models import (
     SemiSupervisedHeatmapTrackerMHCRNN,
     SemiSupervisedRegressionTracker,
 )
+from lightning_pose.utils import io as io_utils
 from lightning_pose.utils.io import (
     check_if_semi_supervised,
     get_keypoint_names,
@@ -599,6 +600,7 @@ def compute_metrics_single(
 
     # get keypoint names
     labels_df = pd.read_csv(labels_file, header=[0, 1, 2], index_col=0)
+    labels_df = io_utils.fix_empty_first_row(labels_df)
     keypoint_names = get_keypoint_names(
         cfg, csv_file=labels_file, header_rows=[0, 1, 2])
     # load predictions
--- a/scripts/converters/dlc2lp.py
+++ b/scripts/converters/dlc2lp.py
@@ -6,87 +6,113 @@ import shutil
 import numpy as np
 import pandas as pd
 
-parser = argparse.ArgumentParser()
-parser.add_argument("--dlc_dir", type=str)
-parser.add_argument("--lp_dir", type=str)
-args = parser.parse_args()
-dlc_dir = args.dlc_dir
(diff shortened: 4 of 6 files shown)


scripts/converters/dlc2lp.py
ksikka, line 1: The big unrelated diffs are because I indented everything behind `if __name__ == "__main__":` and github does not format this well.

ksikka:
Do not merge, fixing a couple test failures.

ksikka:
Test issue was totally unrelated. 
It was https://github.com/Zulko/moviepy/issues/1625#issuecomment-973056577
Fixed by:
```
pip uninstall moviepy decorator
pip install moviepy==1.0.3
```

lightning_pose/data/datasets.py
```diff
@@ -20,6 +19,7 @@
     MultiviewHeatmapLabeledExampleDict,
     generate_heatmaps,
 )
+from lightning_pose.utils import io as io_utils
```
themattinthehatt, line 22: can this and the line below be combined?

lightning_pose/utils/fiftyone.py
```diff
@@ -9,6 +9,7 @@
 from tqdm import tqdm
 from typeguard import typechecked
 
+from lightning_pose.utils import io as io_utils
```
themattinthehatt, line 12: same, can consolidate imports

lightning_pose/utils/scripts.py
```diff
@@ -39,6 +39,7 @@
     SemiSupervisedHeatmapTrackerMHCRNN,
     SemiSupervisedRegressionTracker,
 )
+from lightning_pose.utils import io as io_utils
```
themattinthehatt, line 42: same comment

scripts/converters/dlc2lp.py
```diff
@@ -6,87 +6,115 @@
...
-                ("Unnamed: 1_level_0", "Unnamed: 1_level_1", "Unnamed: 1_level_2"), axis=1,
-            )
-            df_tmp2 = df_tmp1.drop(
-                ("Unnamed: 2_level_0", "Unnamed: 2_level_1", "Unnamed: 2_level_2"), axis=1,
-            )
-            df_tmp2.index = new_col
-            df_tmp = df_tmp2
-    except IndexError:
+
+def fix_empty_first_row(df: pd.DataFrame) -> pd.DataFrame:
```
themattinthehatt, line 10: import from utils

themattinthehatt (changes requested):

scripts/converters/dlc2lp.py
ksikka: This script has no dependency on lightning-pose and I was wondering if we wanted to keep it that way.

ksikka: I guess if they are converting files to lightning-pose, they would already have a conda environment with lightning-pose installed. So no issues in taking a lightning-pose dependency.

themattinthehatt: you have a good point, but yeah let's just assume they're already in an LP environment

lightning_pose/data/datasets.py
ksikka: I don't think so. 

Btw, the reason for this odd inconsistency - I could have imported the function directly out of the io module, like we do for all other utils/io.py functions, but when an unfamiliar reader of the code sees a random util function being used, they wonder where it came from?  By importing the util as `io_utils`, then the fn call looks like `io_utils.foo()` and it's clear where foo is coming from.

Other objects like BaseTrackingDataset, HeatmapTracker, suffer less from the above problem because they are things that have a clear categorization: Dataset, Tracker (which we map to models/ in our mind). In contrast, utils/io.py functions seem like loose functions.

Going back to the original comment - i can either
- leave as is
- migrate utils.io functions to the new pattern in the modified files
- revert: just import `fix_empty_first_row` directly into the namespace

scripts/converters/dlc2lp.py
ksikka: done

lightning_pose/data/datasets.py
themattinthehatt: i'd say at least for this file (we can slowly change other files in future PRs) it would be nice to have
```
from lightning_pose.utils import io as io_utils
```
and remove
```
from lightning_pose.utils.io import get_context_img_paths, get_keypoint_names
```

Then you can use `io_utils.get_context_img_paths` and `io_utils.get_keypoint_names` below

themattinthehatt (approved):

PR #235, opened 2025-01-17
Ckpt path glob escape

hummuscience:
I was using [Hydra interpolation ](https://hydra.cc/docs/upgrades/1.0_to_1.1/defaults_list_interpolation/) to generate `model_name` based on other parameters in the config.

When I used it with the losses_to_use parameter, it ended up with `[ ]` in the model name, which broke the globbing for the ckpt file when doing inference. 

Adding a `glob.escape` solved the issue as it escapes all weird characters when globbing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -151,3 +151,4 @@ multirun/
 preds/
 scripts/.ipynb_checkpoints
 tb_logs
+tests/utils/test_cropzoom_data/
--- a/docs/api/lightning_pose.models.backbones.vit_img_encoder.ImageEncoderViT.rst
+++ b/docs/api/lightning_pose.models.backbones.vit_img_encoder.ImageEncoderViT.rst
@@ -1,0 +1,17 @@
+ImageEncoderViT
+===============
+
+.. currentmodule:: lightning_pose.models.backbones.vit_img_encoder
+
+.. autoclass:: ImageEncoderViT
+   :show-inheritance:
+
+   .. rubric:: Methods Summary
+
+   .. autosummary::
+
+      ~ImageEncoderViT.forward
+
+   .. rubric:: Methods Documentation
+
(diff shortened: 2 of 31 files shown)


ksikka:
Hi, thanks for the contribution. May I ask what you were using `model_name` for? I've found it to be less useful than `hydra.run.dir` for organizing models, and was mulling over the idea of removing this field in the future.

The code in the first commit looks fine. The second commit (merge) overwrote your changes though.

hummuscience:
Yeah, I realized the second commit overwrote everything >.< 

I was using it to see different models in tensorboard instead of the date and time. So I could do something like this:
`model_name: ${model.losses_to_use}_${model.model_type}_${data.csv_file}_${training.imgaug}`

But you are right, I could just use the interpolation in `hydra.run.dir` instead to get nice model folders. The issue I was bumping into is if `mode.losses_to_use = ['temporal']`, then the model_name would have those brackets in that would make the globbing for the ckpt fail. 

I think this will also happen if I switch to setting `model.run.dir: outputs/${model.model_type}/${data.csv_file}/${model.losses_to_use}/${training.imgaug}` since `base_path` will contain brackets

ksikka:
> I think this will also happen if I switch to setting model.run.dir:

Yep. It's also not great that your filename will contain quotes.  If you're only using 1 loss, then you can use `${model.losses_to_use[0]}`. Alternatively add a custom resolver that converts a list to a string by joining the list by comma. I.e call it fmt_list, the you could do `${fmt_list: model.losses_to_use}`. See lightning_pose/__init__.py for an example (LP_ROOT_PATH)

ksikka:
Closing this for housekeeping, hope the issue is resolved on your end? If you still want to, feel free to reopen after the 2nd commit issue is fixed.

PR #236, opened 2025-01-17
Added scale augmentation

hummuscience:
So, I did some testing and adding the scale augmentation improved all the models that include images from different sources (where the animals have different sizes). 

I will post a more quantitative analysis when everything is done, but visually things are better *and* the switches are gone! 

I implemented two options. Scaling + DLC and scaling + DLC + horizontal/vertical flips. 

It is getting a bit wild with all the options. I wondered if it makes sense to let the user pass multiple augmentation building blocks and then build the logic around that. So the user passes a list like so: [dlc, topdown, lr, scale]. 

I also noticed that currently the exact settings of the augmentation can't be changed in the config. Maybe it's a bit overkill for now, but I could imagine augmentations being its own part of the config in the future.

### DIFF
--- a/lightning_pose/data/augmentations.py
+++ b/lightning_pose/data/augmentations.py
@@ -125,7 +125,7 @@ def imgaug_transform(cfg: DictConfig) -> iaa.Sequential:
         ))
 
     else:
-        raise NotImplementedError("must choose imgaug kind from 'default', 'dlc', 'dlc-top-down', 'dlc-scale', 'dlc-top-down-scale'")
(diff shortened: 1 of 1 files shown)


themattinthehatt:
cool @hummuscience, thanks for testing this out. I agree that the options are getting a bit out of hand. let me talk with @ksikka and figure out what a good long-term solution is (I like your idea of a list). We can also figure out how to allow users to change augmentation parameters easily. Previously we had hidden all these options so as not to overwhelm people, but it makes sense that more advanced users will want control over these.

themattinthehatt:
closing this PR as it is subsumed by #259 - @hummuscience please take a look!
Updated docs for custom imgaug pipelines: https://lightning-pose.readthedocs.io/en/latest/source/user_guide_advanced/custom_imgaug_pipeline.html

PR #237, opened 2025-01-18
Add docs on Model, litpose

ksikka:


.readthedocs.yaml
```diff
@@ -9,7 +9,7 @@ version: 2
 build:
   os: ubuntu-20.04
   tools:
-    python: "3.8"
+    python: "3.10"
```
ksikka, line 12: There was a build error with a type annotation that was `list[str]`. Upgraded python to 3.10 to fix.

ksikka: Still some issue, .. autoclass directive not working

ksikka: Nvm, it's working, no idea what was wrong earlier.

themattinthehatt (approved):

PR #242, opened 2025-02-04
Add option to use steps instead of epochs

ksikka:
~~Status: unit tests pass, but I have not done any manual testing.~~
~~I am hoping we can use the pipeline to test, I will also think of how to add unittest coverage for some of these.~~

Update(2/9): added tests. This is a new kind of test that trains a model and then asserts things about the training process using the logs. We test that it trains to the correct number of steps/epochs, unfreezes backbone, anneals LR according to milestones.


ksikka:
@themattinthehatt ready for review. A few callouts:
* backbone unfreezing with steps is faster since the warm up ratio is applied to learning rate per step not per epoch. The warm up ratio is currently not configurable.
* I found an issue with limit_train_batches not doing what we think. see `calculate_train_batches`

Some misc changes in this PR:
* Fix bug in model.py: output_dir argument was not converted to Path, causing methods to break with strings
* train now accepts a model_dir argument, user no longer needs to chdir (note-to-self, update docs in this PR)
* train also allows you to skip evaluation, I added this for tests.
* If there's not enough data, compute_metrics will no longer fail when attempting to compute PCA (for singleview)

themattinthehatt (changes requested):
need to figure out limit_train_batches issue before merging

ksikka:
> need to figure out limit_train_batches issue before merging

Fixed the issue. Previously I was trying to use `calculate_train_batches` as the number of steps per epoch. This returned a value of `10` for small train_frames. This is a limit, and it only applies for unsupervised learning. Using it for supervised doesn't make sense.

I've replaced it with a calculation of steps per epoch. See the most recent commit for these code changes.

Additional cleanups in the last commit: 
- Remove the need to call .setup() on a datamodule. the logic now is entirely triggered from calling the constructor.


One caveat with calculate_steps_per_epoch: we cannot precisely calculate steps per epoch in the rare case of small train_frames, and very short training video (small enough that the 10 step minimum would trigger). our calculation would return 10 steps per epoch, when reality is less. This is rare enough that we can ignore it for now.

tests/utils/test_scripts.py
```diff
@@ -72,59 +97,48 @@ def _unsupervised_multi_gpu_cfg(cfg):
...
-def test_get_data_module_num_gpus_0(cfg):
+def test_get_data_module_num_gpus_0(cfg, mocker):
     cfg = _supervised_multi_gpu_cfg(cfg)
     # when num_gpus is set to 0, i.e. from a deprecated config
     cfg.training.num_gpus = 0
-    data_module = get_data_module(cfg, Mock(spec=BaseTrackingDataset))
+    mock_data_module_init = mocker.patch.object(BaseDataModule, '__init__', return_value=None)
+    get_data_module(cfg, Mock(spec=BaseTrackingDataset))
+
+    get_data_module(cfg, Mock(spec=BaseTrackingDataset))
```
themattinthehatt, line 107: Can remove both `get_data_module` calls, they don't appear to be used.

Can you say more about the motivation to use this mocker pattern? Why not just create a data module?

tests/utils/test_scripts.py
```diff
@@ -72,59 +97,48 @@ def _unsupervised_multi_gpu_cfg(cfg):
...
+    assert mock_data_module_init.call_args.kwargs["val_batch_size"] == cfg.training.val_batch_size
+    assert mock_data_module_init.call_args.kwargs["test_batch_size"] == cfg.training.test_batch_size
 
 
-def test_get_data_module_multi_gpu_batch_size_adjustment_supervised(cfg):
+def test_get_data_module_multi_gpu_batch_size_adjustment_supervised(cfg, mocker):
     cfg = _supervised_multi_gpu_cfg(cfg)
-    data_module = get_data_module(cfg, Mock(spec=BaseTrackingDataset))
+    mock_data_module_init = mocker.patch.object(BaseDataModule, '__init__', return_value=None)
+    get_data_module(cfg, Mock(spec=BaseTrackingDataset))
```
themattinthehatt, line 120: same question - unused?

tests/utils/test_scripts.py
```diff
@@ -72,59 +97,48 @@ def _unsupervised_multi_gpu_cfg(cfg):
...
 
 
 def test_get_data_module_multi_gpu_batch_size_adjustment_unsupervised(
-    cfg, heatmap_dataset, toy_data_dir
+    cfg, heatmap_dataset, toy_data_dir, mocker
 ):
     cfg = _unsupervised_multi_gpu_cfg(cfg)
-    data_module = get_data_module(
+    mock_data_module_init = mocker.patch.object(UnlabeledDataModule, '__init__', return_value=None)
+    get_data_module(
```
themattinthehatt, line 132: unused?

tests/tblogs/test_lr.py
themattinthehatt, line 1: rename to `tests/test_training/test_lr.py`

tests/tblogs/test_train_duration.py
themattinthehatt, line 1: rename to `tests/test_training/test_train_duration.py`

setup.py
```diff
@@ -75,7 +75,8 @@ def get_cuda_version():
     "scikit-learn>=1.5.0",
     "seaborn",
     "streamlit",
-    "tensorboard",
+    "tensorboard<3.0.0",
+    "tbparse<1.0.0",
```
themattinthehatt, line 79: move to extras_require/dev below

themattinthehatt (changes requested):

tests/utils/test_scripts.py
ksikka: One is duplicative, removed.

The get_data_module has the side effect of calling the BaseDataModule constructor, which we then test assertions about.  But we don't need the returned object for this, we use the mocker object to assert what it was called with.

Reason for mocker: Previous code was running the real constructor, with a mock dataset as an input argument. This led to errors once the constructor started calling the _setup method. Mock dataset returns None for all attributes. The specific failure was about dataset.__len__ when the datamodule tries to split the dataset by length.

What the mocker does: replace data module constructor with a fake that does nothing. It just intercepts the args and allows you to assert things about them. This is the ideal thing for a "unit test". It tests just the logic in scripts.py, and exercises none of the code in datamodule.

tests/utils/test_scripts.py
ksikka: Side effect of calling data module constructor

setup.py
ksikka: oops yeah, done.

tests/tblogs/test_train_duration.py
ksikka: done

tests/tblogs/test_lr.py
ksikka: done

tests/utils/test_scripts.py
ksikka: side effect of calling constructor

PR #243, opened 2025-02-09
Automatically set num_workers to cpu count

ksikka:
No longer do you need to explicitly configure num_workers, but you still can if you want


themattinthehatt (approved):
love a good config parameter removal :)

PR #247, opened 2025-02-23
Fix multiview pixelerror computation

ksikka:


ksikka:
Checkout the axon continuous integration passing in this PR! Right now the self-hosted runner is running on my home PC - not a sustainable strategy - but it proves the concept.

themattinthehatt (approved):

PR #250, opened 2025-02-23
Add support for cropzoom pipeline

ksikka:
I have refactored it so that there is no cropzoom in the core lightning_pose code. All cropzoom stuff is invoked externally using `litpose crop` and `litpose remap`.

Reasons for this:
- allow us to modify cropzoom things without risk of breaking core lightning pose code
- if something fails, we can re-run just that thing (instead of all of training, or all of prediction)
- train, predict, crop, remap commands can be farmed out to different kinds of machines using slurm or lightning jobs.

See docs: https://lightning-pose--250.org.readthedocs.build/en/250/source/user_guide_advanced/cropzoom_pipeline.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -143,12 +143,14 @@ data/mirror-mouse-example_split/
 data/Chickadee
 
 # random other outputs that might end up in the repo
+lightning_logs/
 grid_artifacts/
 .vscode/
 outputs/
 multirun/
 preds/
 scripts/.ipynb_checkpoints
+tb_logs
 tests/utils/test_cropzoom_data/
 tests/test_model_mirror_mouse/
 tests/test_model_mirror_mouse_multiview/
--- a/.gitignore
+++ b/.gitignore
@@ -143,14 +143,12 @@ data/mirror-mouse-example_split/
 data/Chickadee
 
 # random other outputs that might end up in the repo
-lightning_logs/
 grid_artifacts/
 .vscode/
 outputs/
 multirun/
 preds/
 scripts/.ipynb_checkpoints
-tb_logs
 tests/utils/test_cropzoom_data/
 tests/test_model_mirror_mouse/
 tests/test_model_mirror_mouse_multiview/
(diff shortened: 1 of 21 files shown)


.gitignore
themattinthehatt, line 1: why remove these directories?

cli/main.py
```diff
@@ -5,8 +5,11 @@
 import os
 import sys
 from pathlib import Path
+from textwrap import dedent
 from typing import TYPE_CHECKING
 
+from omegaconf import OmegaConf
+
 if TYPE_CHECKING:
```
themattinthehatt, line 13: can you write a short comment about what this is doing?

lightning_pose/train.py
```diff
@@ -63,7 +64,7 @@ def train(
 
     if not skip_evaluation:
         _evaluate_on_training_dataset(model)
-        _evaluate_on_ood_dataset(model)
+        _evaluate_on_training_dataset(model, ood_mode=True)
```
themattinthehatt, line 67: nice

tests/conftest.py
```diff
@@ -531,27 +531,19 @@ def trainer(cfg) -> pl.Trainer:
         check_val_every_n_epoch=1,
         log_every_n_steps=1,
         callbacks=callbacks,
+        enable_checkpointing = False,
         limit_train_batches=2,
         num_sanity_val_steps=0,
+        logger=False,
```
themattinthehatt, line 537: :ok_hand:

.gitignore
```diff
@@ -143,14 +143,12 @@ data/mirror-mouse-example_split/
...
 
 # random other outputs that might end up in the repo
-lightning_logs/
 grid_artifacts/
 .vscode/
 outputs/
 multirun/
 preds/
 scripts/.ipynb_checkpoints
-tb_logs
```
themattinthehatt, line 153: should keep these around just in case

themattinthehatt (changes requested):

PR #259, opened 2025-03-03
imgaug refactor

themattinthehatt:
The current imgaug code defines a handful of hard-coded pipelines that are defined by strings like "dlc", "dlc-top-down", etc. While succinct, this approach is not flexible and does not scale well.

To address this issue, I have refactored the imgaug code so that it can continue to use the previous string aliases (i.e. is backwards compatible), but also allows users to define their own custom pipeline in the config file.

Specifically:
- `lightning_pose.data.augmentations.imgaug_transform` is now a fully flexible core function that takes in a parameter dict and constructs an imgaug pipeline; the parameter dict defines the types of augmentations used, their various parameters, and their order. Unlike before, it does not rely on the bulky `cfg` DictConfig that specfies the entire Lightning Pose pipeline (data, model, training, etc)
- `lightning_pose.utils.scripts.get_imgaug_transform` now does more of the heavy lifting; it hard-codes the pipeline details if a user selects one of the previously used strings, but also allows for yaml files to define a custom pipeline.

generalizes #236 

This PR also removes the Resize augmentation from the end of the imgaug pipeline, and instead makes this a more flexible option that datasets can add in their constructors. This will be useful for more complex (i.e. 3d) image augmentation pipelines.

### DIFF
--- a/tests/data/test_datamodules.py
+++ b/tests/data/test_datamodules.py
@@ -1,5 +1,6 @@
 """Test datamodule functionality."""
 
+import numpy as np
 import pytest
 import torch
 from lightning.pytorch.utilities import CombinedLoader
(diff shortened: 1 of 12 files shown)


tests/data/test_datamodules.py
```diff
@@ -71,20 +90,58 @@ def test_multiview_heatmap_datamodule(cfg_multiview, multiview_heatmap_data_modu
...
+    # check imgaug pipeline makes repeatable data
+    b1 = multiview_heatmap_data_module.test_dataset[0]
+    b2 = multiview_heatmap_data_module.test_dataset[0]
+    assert np.allclose(b1["images"], b2["images"])
+    assert np.allclose(b1["keypoints"], b2["keypoints"], equal_nan=True)
 
     # cleanup
     del batch
+    del b1
+    del b2
```
ksikka, line 144: Just FYI these dels aren't necessary. del the variable reference to the data. all function-scoped variables will anyways be cleaned up when the function exits

ksikka (approved):
Looks good! I extracted out a function `expand_imgaug_str_to_dict`,  ran `black` on augmentations.py, and fixed a typo. See commit history.

tests/data/test_datamodules.py
ksikka: Removed dels in this file. I just noticed there were some calls to empty torch cache. I think the GPU footprint of stuff in this file should be minimal compared to model weights. After these tests are done, all references to the data will be gone, and it can be cleaned up by garbage collection at a later time.

In general we could run torch clear cache before training or predicting a model. It's cleaner than at the end, since we don't have to `del` a bunch of variables, they're already gone due to functions exiting.

PR #260, opened 2025-03-03
Fix multigpu test

ksikka:


themattinthehatt:
thanks for looking into this - any idea why it was working previously?

ksikka:
Of course. I recently tested that multigpu tests pass in axon, then updated the `srun` to have gres:2 to allow multigpu tests to run in CI. 

However sometime after that, I must have run `pip install lightning-pose` from pypi, not using -e, and this breaks the multigpu tests as the subprocess call to `python script/train_hydra.py` resolves `import lightning_pose` from the pip installed lightning pose, not the root of the git repo that you run `pytest` in. Pytest uses the prepend method described here, but it does not extend to subprocesses:
https://docs.pytest.org/en/7.1.x/explanation/pythonpath.html

PR #279, opened 2025-04-17
head refactoring

themattinthehatt:
This refactor significantly cleans up the model code by creating "heads" that process the outputs of the backbones, allowing that potential complexity to isolated inside its own object.

I have verified that the refactor produces the exact same training curves for the following models on the mirror-mouse dataset:
* heatmap
* heatmap_mhcrnn
* heatmap (mutliview)
* heatmap_mhcrnn (multiview)

I also checked the regression models, and while the curves weren't exactly the same they were extremely close (not sure what's causing the discrepancy, given the heatmap models matched exactly, but I'm not so concerned).


lightning_pose/callbacks.py
```diff
@@ -73,10 +73,6 @@ def __init__(
         self._warmed_up = False
 
     def on_train_batch_start(self, trainer, pl_module, batch, batch_idx):
-        # This callback is only applicable to heatmap models but we
-        # might encounter a RegressionModel.
-        if not hasattr(pl_module, "upsampling_layers"):
-            return
```
ksikka, line 79: I wonder if the removal of this logic is why regression loss curve looks different? 
Should this logic be replaced?

lightning_pose/models/regression_tracker.py
```diff
@@ -139,6 +137,13 @@ def predict_step(
         confidence = torch.zeros((predicted_keypoints.shape[0], predicted_keypoints.shape[1] // 2))
         return predicted_keypoints, confidence
 
+    def get_parameters(self):
+        params = [
+            {"params": self.backbone.parameters(), "lr": 0, "name": "backbone"},
+            {"params": self.head.parameters(), "name": "head"},
+        ]
+        return params
```
ksikka, line 145: Another hint that the regression loss curve discrepancy is due to the backbone unfreezing (prior to this PR, I don't think it was doing backbone unfreezing).

lightning_pose/models/base.py
```diff
@@ -454,14 +454,7 @@ def get_scheduler(self, optimizer):
...
     def get_parameters(self):
-        if getattr(self, "upsampling_layers", None) is not None:
-            params = [
-                {"params": self.backbone.parameters(), "lr": 0, "name": "backbone"},
-                {"params": self.upsampling_layers.parameters(), "name": "upsampling"},
-            ]
-        else:
-            params = filter(lambda p: p.requires_grad, self.parameters())
-
+        params = filter(lambda p: p.requires_grad, self.parameters())
```
ksikka, line 457: Just curious, is `get_parameters` overridden in all cases now?

lightning_pose/utils/scripts.py
```diff
@@ -455,7 +452,6 @@ def get_model(
                 backbone=cfg.model.backbone,
                 pretrained=backbone_pretrained,
                 downsample_factor=cfg.data.get("downsample_factor", 2),
-                output_shape=data_module.dataset.output_shape,
```
ksikka, line 458: Just curious what enabled the removal of output_shape?

ksikka (approved):

ksikka:
Nice!

lightning_pose/models/base.py
themattinthehatt: I believe so. We'll probably continue to override this in all future models, but figured I'd keep this here for now as a fallback (though maybe unnecessary?)

lightning_pose/utils/scripts.py
themattinthehatt: Ha I searched the repo and it was never used anywhere. This value was set using the `output_shape` attribute of the data loader, which does get used throughout the repo, but it was never accessed from the model itself. My guess is that this is an artifact of a much earlier implementation.

PR #283, opened 2025-04-22
Pick best checkpoint by default

hummuscience:
This PR addresses this issue: https://github.com/paninski-lab/lightning-pose/issues/282

I think that eventually it should be possible to choose the checkpoint to use (I think DLC also allows the as an option).

But it might be too specific of a use case for now. So choosing the one marked with "best" is the easiest.

### DIFF
--- a/lightning_pose/utils/io.py
+++ b/lightning_pose/utils/io.py
@@ -113,7 +113,6 @@ def ckpt_path_from_base_path(
             # Try to find the one with the highest step count.
             max_step = -1
             latest_ckpt = None
-            parse_errors = 0
 
             for f in latest_version_files:
                 match = re.search(r"step=(\d+)", f)
@@ -122,19 +121,16 @@ def ckpt_path_from_base_path(
                     if step > max_step:
                         max_step = step
                         latest_ckpt = f
-                else:
-                    parse_errors += 1
 
             if latest_ckpt is not None:
                 return latest_ckpt
-            elif parse_errors == len(latest_version_files):
-                # Could not parse step count from any file, return the lexicographically last one as a fallback
-                warnings.warn("Could not parse step counts from any checkpoint. Returning the lexicographically last file.")
-                return sorted(latest_version_files)[-1]
             else:
-                # Should not happen if latest_version_files is not empty
-                warnings.warn("Unexpected state: Multiple checkpoints, none best, step parsing partially successful.")
-                return latest_ckpt if latest_ckpt else sorted(latest_version_files)[-1]
+                # Could not determine which checkpoint to use
+                raise ValueError(
+                    f"Multiple checkpoint files found but cannot determine which to use: {latest_version_files}. "
+                    "None are marked as 'best' and cannot parse step counts to determine latest. "
+                    "Please manually select the appropriate checkpoint."
+                )
             
 @typechecked
 def check_if_semi_supervised(losses_to_use: ListConfig | list | None = None) -> bool:


hummuscience:
I just realized it also includes the `glob.escape` fix that I sent a PR for a while ago. Feel free to remove it as its something that probably only affects me in this case. 

Otherwise, let me know if I should push a fix for this so keep the history clean

lightning_pose/utils/io.py
```diff
@@ -32,51 +32,86 @@ def ckpt_path_from_base_path(
...
+    latest_version_files = ckpt_files_by_version[latest_version]
 
-    latest_version = max(ckpt_file_by_version.keys())
-    return ckpt_file_by_version[latest_version]
+    # Find the "best" checkpoint in the latest version
+    best_ckpt_file = None
+    for f in latest_version_files:
+        if "-best.ckpt" in os.path.basename(f):
+            best_ckpt_file = f
+            break  # Found the best file, stop searching
```
ksikka, line 94: optional stylistic comment: you can directly return best_ckpt_file here (instead of break and return later)

nice use of comments btw, I found it very easy to understand, thanks

lightning_pose/utils/io.py
```diff
@@ -32,51 +32,86 @@ def ckpt_path_from_base_path(
...
+    # Get the latest version's checkpoint files
+    latest_version = max(ckpt_files_by_version.keys())
+    latest_version_files = ckpt_files_by_version[latest_version]
 
-    latest_version = max(ckpt_file_by_version.keys())
-    return ckpt_file_by_version[latest_version]
+    # Find the "best" checkpoint in the latest version
+    best_ckpt_file = None
+    for f in latest_version_files:
+        if "-best.ckpt" in os.path.basename(f):
```
ksikka, line 92: optional stylistic comment: f.endswith("-best.ckpt")

lightning_pose/utils/io.py
```diff
@@ -32,51 +32,86 @@ def ckpt_path_from_base_path(
...
+            best_ckpt_file = f
+            break  # Found the best file, stop searching
 
+    if best_ckpt_file:
+        # Found the 'best' checkpoint
+        return best_ckpt_file
+    else:
+        # No 'best' checkpoint found
```
ksikka, line 100: In what scenario would there be no "best" checkpoint? A model that was trained for too few epochs? 

Since this is an unusual scenario, maybe we should emit a warning
"No 'best' checkpoint found, falling back to latest checkpoint'.

lightning_pose/utils/io.py
ksikka, line 94: I think multiple best checkpoints are possible if someone adds the `save_top_k` parameter to the ModelCheckpoint callback. In this case we should still raise an error about not being able to select from multiple checkpoints, instead of returning the first one found.

ksikka:
Thanks for the fix, I did not realize people used `ckpt_every_n_epochs`. I am glad you found and fixed the issue. 

OK to leave the glob escape.
Nice use of comments btw, I found it very easy to understand, thanks.

lightning_pose/utils/io.py
hummuscience: Yeah. Technically that should "never" happen, right? Even if one interrupts the training. Does it fall back to the latest model being the "best"?

lightning_pose/utils/io.py
hummuscience: Should I implement this possibility or just output an error/warning for now?

ksikka: I see you've implemented it

lightning_pose/utils/io.py
```diff
@@ -32,52 +33,107 @@ def ckpt_path_from_base_path(
...
+            ckpt_step_counts = {}
+            max_step = -1
+            latest_ckpt = None
+            parse_errors = 0
+
+            for f in latest_version_files:
+                match = re.search(r"step=(\d+)", f)
+                if match:
+                    step = int(match.group(1))
+                    ckpt_step_counts[f] = step
```
ksikka, line 118: This dict is unused and should be removed

lightning_pose/utils/io.py
```diff
@@ -32,52 +33,107 @@ def ckpt_path_from_base_path(
...
+                    if step > max_step:
+                        max_step = step
+                        latest_ckpt = f
+                else:
+                    parse_errors += 1
+                    warnings.warn(f"Could not parse step count from checkpoint file: {f}")
+
+            if latest_ckpt is not None:
+                return latest_ckpt
+            elif parse_errors == len(latest_version_files):
```
ksikka, line 128: I'd remove any logic based on parse errors, it's being needlessly cautious IMO. If we're in this scenario it needs manual investigation.

ksikka:
A couple of pending comments. Note on the test failure, it seems totally unrelated. I will have to investigate asynchronously.

ksikka:
Hey @hummuscience ugh I just saw that this has been awaiting my review for months now. I want to get this merged as smoothly as possible for you. Investigating if I can re-run tests on top of latest main...

ksikka:
Tests pass after rebase, which means this probably doesn't break anything. (We invoke this logic in model api tests).  
So I'll merge this and see if I can easily add unit tests after.

Thanks @hummuscience !

PR #284, opened 2025-04-22
Added support for random weighted sampling for unbalanced datasets

hummuscience:
This is a draft pull request for adding the weighted random sampling option as suggested here: https://github.com/paninski-lab/lightning-pose/issues/158#issuecomment-2299752505

I am not very experienced with coding and good practices (typical biologist background :D) so I put this together with the help of an LLM.

For some reason, I was unable to pass the config setting to the function. Am I missing something? This is the reason it's currently set to be enabled by default. Technically, even when its turned on, it should still work normally for data that is not unbalanced, right?

I also added a test to check if the functionality works. Even though, it makes more sense to add a test with an unbalanced dataset as input and check that the outputs are correct. Right?

Also, I am not sure yet how well this works with the suggestion to use the COCO input for heterogenous datasets: https://github.com/paninski-lab/lightning-pose/issues/263

As I said, not much experience here and would love some input on how to do this right. 

I am also open for a meeting (as @themattinthehatt suggested) to discuss this and also adding the top_view_mouse model to LP.

### DIFF
--- a/lightning_pose/data/datamodules.py
+++ b/lightning_pose/data/datamodules.py
@@ -9,7 +9,7 @@ import lightning.pytorch as pl
 import torch
 from lightning.pytorch.utilities import CombinedLoader
 from omegaconf import DictConfig
-from torch.utils.data import DataLoader, Subset, random_split
+from torch.utils.data import DataLoader, Subset, random_split, WeightedRandomSampler, RandomSampler
 
 from lightning_pose.data.dali import PrepareDALI
 from lightning_pose.data.datatypes import SemiSupervisedDataLoaderDict
@@ -38,6 +38,7 @@ class BaseDataModule(pl.LightningDataModule):
         test_probability: float | None = None,
         train_frames: float | int | None = None,
         torch_seed: int = 42,
+        enable_weighted_sampler: bool = True,
     ) -> None:
         """Data module splits a dataset into train, val, and test data loaders.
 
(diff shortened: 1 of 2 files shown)


themattinthehatt:
Thanks for the PR @hummuscience! Happy to take a closer look soon. I'm a bit swamped until mid-May with end of semester/deadlines, but after that let's definitely plan to meet and discuss further (both this PR and the top_view_mouse model). This work will actually dovetail quite nicely with the COCO input for heterogeneous datasets issue.

themattinthehatt:
@hummuscience you were so far ahead of me on this one! Sorry I left this lingering for so long, I got very in-the-weeds with the Lightning Pose 3D project. I'm finally circling back to this problem of training a "super animal" model across multiple datasets. I have a couple PRs that are building out the infrastructure better:

- #440 : more formally separating the "occluded" condition from "not labeled"
- #446 : properly implementing the horizontal flip augmentation that caused you so many headaches

Right now I'm playing around with some datasets where I've subsampled each to have the same number of frames, but once I have some proof-of-concept results with that I'll want to come back to this PR.

hummuscience:
Funnily, I ended up cleaning this up with a slightly different approach to make this work :D I could have a look how much it diverged from what I did here

themattinthehatt:
Please let me know, I'm curious! I'll still be playing around with artificially balanced datasets for the next couple of weeks but will then look to move beyond that and fit models on the full, imbalanced datasets.

PR #287, opened 2025-04-24
bug fix in left-right/top-down augmentation pipeline

themattinthehatt:
Fix to a pretty serious bug in the `dlc-lr` and `dlc-top-down` image augmentation pipelines - I was erroneously flipping the images/keypoints rather than rotating them. 

For example, imagine a top-down view of an animal that is pointing towards the top of the frame, with the center vertical axis of the frame running down the axis of the animal. If the left/right ears are labeled as such, flipping over the vertical axis would result in an animal that looks almost the same, but the left/right ear keypoints are now on the wrong side of the head.

Instead of flipping images we should be rotating them.

### DIFF
--- a/lightning_pose/data/augmentations.py
+++ b/lightning_pose/data/augmentations.py
@@ -63,11 +63,15 @@ def imgaug_transform(params_dict: dict | DictConfig) -> iaa.Sequential:
         transform_args = args.get("args", ())
         transform_kwargs = args.get("kwargs", {})
 
-        # make sure any lists are converted to tuples; DictConfig cannot load tuples from yaml
-        # files, but no iaa args are lists
+        # DictConfig cannot load tuples from yaml files
+        # make sure any lists are converted to tuples
+        # unless the list contains a single item, then pass through the item (hack for Rot90)
         for kw, arg in transform_kwargs.items():
             if isinstance(arg, list) or isinstance(arg, ListConfig):
-                transform_kwargs[kw] = tuple(arg)
+                if len(arg) == 1:
+                    transform_kwargs[kw] = arg[0]
+                else:
+                    transform_kwargs[kw] = tuple(arg)
 
         # add transform to pipeline
         if apply_prob == 0.0:
@@ -91,13 +95,13 @@ def expand_imgaug_str_to_dict(params: str) -> dict[str, Any]:
         pass  # no augmentations
     elif params in ["dlc", "dlc-lr", "dlc-top-down"]:
 
-        # flip horizontally
-        if params in ["dlc-lr", "dlc-top-down"]:
-            params_dict["Fliplr"] = {"p": 1.0, "kwargs": {"p": 0.5}}
+        # rotate 0 or 180 degrees
+        if params in ["dlc-lr"]:
+            params_dict["Rot90"] = {"p": 1.0, "kwargs": {"k": [[0, 2]]}}
 
-        # flip vertically
+        # rotate 0, 90, 180, or 270 degrees
         if params in ["dlc-top-down"]:
-            params_dict["Flipud"] = {"p": 1.0, "kwargs": {"p": 0.5}}
+            params_dict["Rot90"] = {"p": 1.0, "kwargs": {"k": [[0, 1, 2, 3]]}}
 
         # rotate
         rotation = 25  # rotation uniformly sampled from (-rotation, +rotation)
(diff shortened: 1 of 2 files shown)


hummuscience:
Oh wow.... That might explain, a lot!

hummuscience:
I was trying to understand why this was wrong, but failed (maybe I am missing something). 

DLC are doing something simiar: https://github.com/DeepLabCut/DeepLabCut/blob/a921c5bc01e8a8a0a26cf649794c998b3a3cda0f/deeplabcut/pose_estimation_tensorflow/datasets/pose_imgaug.py#L34

themattinthehatt:
I definitely believe this is an explanation for your previous results, particularly the issue you were seeing with side collapse. I'll try to draw up a little diagram to illustrate it.

DLC are indeed doing something similar, I more or less copied this pipeline initially. It's actually not a problem if all the keypoints lie on the midline of the body. There might be some other cases where "mirror" isn't a bad idea. But for a general top-down view it is not going to work, so not sure what they did there.

hummuscience:
Fliplr is set to false by default as it is not label-preserving. That really could explain a lot of what I am seeing... 

https://deeplabcut.github.io/DeepLabCut/docs/recipes/pose_cfg_file_breakdown.html#fliplr-or-a-horizontal-flip

lightning_pose/data/augmentations.py
```diff
@@ -63,11 +63,15 @@ def imgaug_transform(params_dict: dict | DictConfig) -> iaa.Sequential:
...
+        # DictConfig cannot load tuples from yaml files
+        # make sure any lists are converted to tuples
+        # unless the list contains a single item, then pass through the item (hack for Rot90)
         for kw, arg in transform_kwargs.items():
             if isinstance(arg, list) or isinstance(arg, ListConfig):
-                transform_kwargs[kw] = tuple(arg)
+                if len(arg) == 1:
+                    transform_kwargs[kw] = arg[0]
+                else:
+                    transform_kwargs[kw] = tuple(arg)
```
hummuscience, line 74: This is probably due to this line

lightning_pose/data/augmentations.py
```diff
@@ -91,13 +95,13 @@ def expand_imgaug_str_to_dict(params: str) -> dict[str, Any]:
...
-            params_dict["Fliplr"] = {"p": 1.0, "kwargs": {"p": 0.5}}
+        # rotate 0 or 180 degrees
+        if params in ["dlc-lr"]:
+            params_dict["Rot90"] = {"p": 1.0, "kwargs": {"k": [[0, 2]]}}
 
-        # flip vertically
+        # rotate 0, 90, 180, or 270 degrees
         if params in ["dlc-top-down"]:
-            params_dict["Flipud"] = {"p": 1.0, "kwargs": {"p": 0.5}}
+            params_dict["Rot90"] = {"p": 1.0, "kwargs": {"k": [[0, 1, 2, 3]]}}
```
hummuscience, line 104: It is actually valid to pass a list to Rot90. But it means that this will fail due to the above forcing of a tuple

hummuscience (commented):
I tried to mimic the Rotation (and leave out the flips) in the config file like so:

```
  imgaug:
    Affine:
      p: 1
      kwargs:
        scale: [0.5, 1.5]
##    Fliplr:
##      p: 1
##      kwargs:
##        p: 0.5
##    Flipud:
##      p: 1
##      kwargs:
##        p: 0.5
    Rot90:
      p: 1.0
      kwargs:
        k: [0,1,2,3]
    Rotate:
      p: 0.4
      kwargs:
        rotate: [-25, 25]
    MotionBlur:
      p: 0.5
      kwargs:
        k: 5
        angle: [-90, 90]
    CoarseDropout:
      p: 0.5
      kwargs:
        p: 0.02
        size_percent: 0.3
        per_channel: 0.5
    CoarseSalt:
      p: 0.5
      kwargs:
        p: 0.01
        size_percent: [0.05, 0.1]
    CoarsePepper:
      p: 0.5
      kwargs:
        p: 0.01
        size_percent: [0.05, 0.1]
    ElasticTransformation:
      p: 0.5
      kwargs:
        alpha: [0, 10]
        sigma: 5
    AllChannelsHistogramEqualization:
      p: 0.1
    AllChannelsCLAHE:
      p: 0.1
    Emboss:
      p: 0.1
      kwargs:
        alpha: [0, 0.5]
        strength: [0.5, 1.5]
    CropAndPad:
      p: 0.4
      kwargs:
        percent: [-0.15, 0.15]
        keep_size: false
```


But I am getting an error that k was expected to be a tuple with two arguments. Instead, it had 4.

PR #294, opened 2025-06-03
fix vit_b_sam predict model init issue

PPWangyc:
Fix the vit-b-sam model init issue in `def get_model` in 

> utils/scripts.py


PPWangyc:
This pull request solves this issue: https://github.com/paninski-lab/lightning-pose/issues/261

PR #296, opened 2025-06-04
keep lists longer than 2 as lists

hummuscience:
Deals with this issue:

https://github.com/paninski-lab/lightning-pose/issues/295

I am not sure if this is the best approach. Will it fail with other imgaug parameters? For now it fixed my error.

An alternative would be to make this specific for Rot90 (for now)

### DIFF
--- a/lightning_pose/data/augmentations.py
+++ b/lightning_pose/data/augmentations.py
@@ -57,7 +57,6 @@ def imgaug_transform(params_dict: dict | DictConfig) -> iaa.Sequential:
     data_transform = []
 
     for transform_str, args in params_dict.items():
-
         transform = getattr(iaa, transform_str)
         apply_prob = args.get("p", 0.5)
         transform_args = args.get("args", ())
@@ -70,8 +69,10 @@ def imgaug_transform(params_dict: dict | DictConfig) -> iaa.Sequential:
             if isinstance(arg, list) or isinstance(arg, ListConfig):
                 if len(arg) == 1:
                     transform_kwargs[kw] = arg[0]
-                else:
+                elif len(arg) == 2:
                     transform_kwargs[kw] = tuple(arg)
+                else:
+                    transform_kwargs[kw] = arg
 
         # add transform to pipeline
         if apply_prob == 0.0:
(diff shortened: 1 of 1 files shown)


lightning_pose/data/augmentations.py
themattinthehatt, line 1: A few requests to wrap this up:

1. can you please add this text to the Examples?
```
        Create a pipeline with
        - Rot90 transformation that is applied 100% of the time with rotations of 0, 90, 180, or 270 degrees.
        >>> params_dict = {
        >>>     'Rot90': {'p': 1.0, 'kwargs': {'k': [[0, 1, 2, 3]]}},  # note the (required) nested list
        >>> }

        In a config file, this will look like:
        >>> training:
        >>>   imgaug:
        >>>     Rot90:
        >>>       p: 1.0
        >>>       kwargs:
        >>>         k: [0, 1, 2, 3]

       NOTE: due to the way yaml files are parsed, Rot90 cannot take a list of 2 values, otherwise it is parsed as a
       tuple and all rotations between the two values will be sampled uniformly.

```

2. Please add these tests to the bottom of `lightning-pose/tests/data/test_augmentations.py` (no need to make a new function, can keep in the single test function that exists there already).

```
params_dict = {"Rot90": {"p": 1.0, "kwargs": {"k": [[0, 1]]}}}
pipe = imgaug_transform(params_dict)
assert pipe.__str__().find("Choice(a=[0, 1]") > -1

params_dict = {"Rot90": {"p": 1.0, "kwargs": {"k": [[0, 1, 4]]}}}
pipe = imgaug_transform(params_dict)
assert pipe.__str__().find("Choice(a=[0, 1, 4]") > -1

params_dict = {"Rot90": {"p": 1.0, "kwargs": {"k": [0, 1, 4]}}}
pipe = imgaug_transform(params_dict)
assert pipe.__str__().find("Choice(a=[0, 1, 4]") > -1

params_dict = {"Rot90": {"p": 1.0, "kwargs": {"k": [0, 3]}}}
pipe = imgaug_transform(params_dict)
assert pipe.__str__().find("parameters=[DiscreteUniform(Deterministic(int 0), Deterministic(int 3)), True]") > -1
```

themattinthehatt (approved):

PR #297, opened 2025-06-17
Check in the app progress so far

ksikka:
Figured it might be a good point to check in the app into the repo, so the changes don't accumulate _too_ much, also so I don't accidentally push -f and overwrite my local changes and then we don't have anything.

Also github says this is 20K lines, but out of that, 17K is the package-lock.json which is a machine-generated package list.


ksikka:
Closing since this is now a separate repo

PR #298, opened 2025-06-19
Add a hidden run_app command

ksikka:


lightning_pose/cli/main.py
```diff
@@ -42,7 +42,3 @@ def main():
 
     # Execute the command
     command_handler(args)
-
-
-if __name__ == "__main__":
```
themattinthehatt, line 47: does the CLI work without this part? How does `main` get called?

ksikka: the pyproject.toml contains
```
[tool.poetry.scripts]
litpose = "lightning_pose.cli.main:main"
```

which will invoke the main function in the main module, so no need to have this if.

with this if block, it was double invoking (one invocation was failing)

PR #302, opened 2025-06-20
new backbone: ViT pretrained on imagenet with mae loss

themattinthehatt:
The LP repo currently only supports the Segment Anything ViT backbone; this PR introduces another backbone (still vit-b/16) but pretrained on imagenet using a masked autoencoding loss (from HuggingFace). This PR opens the door to using our own pretrained ViT models.

### DIFF
--- a/lightning_pose/models/backbones/vit_mae.py
+++ b/lightning_pose/models/backbones/vit_mae.py
@@ -1,0 +1,238 @@
+import math
+
+import torch
+import torch.nn as nn
+from transformers import ViTMAEConfig, ViTMAEForPreTraining
+
+# to ignore imports for sphix-autoapidoc
+__all__ = [
+    "ViTVisionEncoder",
+]
+
+
+class ViTVisionEncoder(nn.Module):
+    """Wrapper around HuggingFace's ViTMAE Vision Encoder."""
+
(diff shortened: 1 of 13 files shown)


lightning_pose/models/backbones/vit_mae.py
```diff
@@ -0,0 +1,238 @@
...
+        )
+        # skip the cls token
+        outputs = outputs[:, 1:, ...]  # [N, S, D]
+        # change the shape to [N, H, W, D] -> [N, D, H, W]
+        S = outputs.shape[1]
+        H, W = math.isqrt(S), math.isqrt(S)
+        outputs = outputs.reshape(N, H, W, -1).permute(0, 3, 1, 2)
+        return outputs
+
+    def _resize_pos_embed(self):
```
PPWangyc, line 122: Do we need to add this new custom function for resizing position embeds? Does the HF transformers automatically resize pos_embed based on different patch size?

themattinthehatt: yeah definitely need this, otherwise there is a size mismatch between the position embeddings and patch embeddings. how did you handle this in your fork? did you resize the images for LP to 224x224?

PR #308, opened 2025-08-14
Calibration plot for frame diagnostics streamlit

hummuscience:
For downstream applications (such as KeypointMoseq) the confidence/likelihood of each prediction are important. 

litpose models often seem to be "over confident" in their predictions. I found myself comparing the models with calibration curves. I tried adding it to the app and it looks okay. 

Here is an example comparing a few models. Not sure if this is useful, let me know. 

newplot(1)


themattinthehatt:
thanks @hummuscience! nice addition. one request, would you mind leaving the heatmap arg in the script? it's not used very often, but occasionally we turn this on for debugging or visualization purposes.

PR #328, opened 2025-10-17
Fix multi view model prediction

### LINKED ISSUE #327: Multi view transformer doesn't run prediction properly
I noticed that after training a `heatmap_multiview_transformer` model, only one video from each view would be processed afterwards.

This is because the `find_video_files_for_views` function only returns one video for each view, and not necessarily even videos that correspond to the same session.

Additionally, it's not possible to run inference for videos from different views at the same time using `litpose predict`

Dominic-DallOsto:
Fixes #327 

I changed the `find_videos_files_for_views` function to return a list of lists of videos     like `[[session0_view0.mp4, session0_view1.mp4, ...], [session1_view0.mp4, session1_view1.mp4, ...], ...]`.

This then lets us loop over all sessions in the training videos directory and feed the videos from all the different views to the model.

Additionally, I modified `litpose predict` for multi view models to accept a list of videos from multiple views, or a list of directories that each contain videos from multiple views. If there's a different way of handling this that you prefer that's no problem!

### DIFF
--- a/tests/utils/test_io.py
+++ b/tests/utils/test_io.py
@@ -311,7 +311,7 @@ def test_split_video_files_per_view():
 
     # Test 1: one session with two views
     video_files_per_session = split_video_files_by_view(
-        [Path("session1_top"), Path("session1_bot")], view_names
+        [Path("/video_dir/session1_top"), Path("/video_dir/session1_bot")], view_names
     )
     assert len(video_files_per_session) == 1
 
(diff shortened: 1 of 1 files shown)


Dominic-DallOsto:
Ahh, the test for `find_video_files_for_views` is failing because I changed its behaviour. Happy to update it if you think the new behaviour makes sense.

ksikka:
Thanks for the feature! I liked the proposal and just made a few minor tweaks as I reviewed. 
https://github.com/Dominic-DallOsto/lightning-pose/pull/1

(My first time reviewing a PR by following up with a PR - I thought this might be the easiest way to minimize the work for you to accept the changes)

Do you mind adding test cases for the new split function?

PR #330, opened 2025-10-22
Add full CLI documentation using sphinx-argparse

ksikka:
This is intended to help cropzoom usage in particular.


ksikka:
Noticed some issues I need to fix before merge

PR #352, opened 2025-12-12
Refactor state dict loading in vits.py

### LINKED ISSUE #351: Positional embedding load missing in pretrained ViT backbone
In this load [checkpoint function,](https://github.com/paninski-lab/lightning-pose/blob/main/lightning_pose/models/backbones/vits.py#L71-L105) we skip the positional embeddings:
We need to remove this part, and use the checkpoint's `position_embeddings`

```
# Skip known problematic layers with size mismatches
if any(prob in model_key for prob in [
    "position_embeddings",
    "patch_embeddings.projection",  # in case backbone was trained with grayscale imgs
    "decoder_pos_embed",
    "decoder_pred",
]):
    continue
```

PPWangyc:
Removed the skip condition for problematic layers in the state dict loading process.

closes #351

### DIFF
--- a/lightning_pose/models/backbones/vits.py
+++ b/lightning_pose/models/backbones/vits.py
@@ -89,14 +89,6 @@ def load_vit_backbone_checkpoint(base, checkpoint: str):
     for key, value in ckpt_vit_pretrain.items():
         if key.startswith("vit_mae."):
             model_key = key.replace("vit_mae.vit.", "")
-            # Skip known problematic layers with size mismatches
-            if any(prob in model_key for prob in [
-                "position_embeddings",
-                "patch_embeddings.projection",  # in case backbone was trained with grayscale imgs
-                "decoder_pos_embed",
-                "decoder_pred",
-            ]):
-                continue
             # Check if shapes match before including in state dict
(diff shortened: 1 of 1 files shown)


lightning_pose/models/backbones/vits.py
```diff
@@ -89,14 +89,6 @@ def load_vit_backbone_checkpoint(base, checkpoint: str):
     for key, value in ckpt_vit_pretrain.items():
         if key.startswith("vit_mae."):
             model_key = key.replace("vit_mae.vit.", "")
-            # Skip known problematic layers with size mismatches
-            if any(prob in model_key for prob in [
-                "position_embeddings",
-                "patch_embeddings.projection",  # in case backbone was trained with grayscale imgs
-                "decoder_pos_embed",
-                "decoder_pred",
-            ]):
```
themattinthehatt: I think we should continue to skip `decoder_pos_embed` and `decoder_pred`, right?

themattinthehatt: I guess the decoder keys won't be in the vision encoder state dict though, so not a problem!

PR #379, opened 2026-02-25
Add Model.predict_frame() for single-frame inference

retconrecon:
## Summary

- Adds `Model.predict_frame(frame_rgb, bbox=None)` for single-frame inference without file I/O or DALI
- Handles standard HeatmapTracker and MHCRNN (context) models
- Includes input validation (dtype, shape, bbox bounds, model type guard)
- Remaps keypoints from resize-space to original frame coordinates
- Adds docs example to `api.rst`

## Tests

- `test_predict_frame` — full-frame inference, shape/dtype/range assertions
- `test_predict_frame_with_bbox` — bbox crop + coordinate remap
- `test_predict_frame_errors` — error-path tests for all validation guards
- `test_predict_frame_bbox_clipping` — bbox extending past frame edge

## Notes

- Preprocessing uses cv2 (not DALI) since DALI cannot process single numpy frames. Docstring documents the divergence and advises against mixing results.
- For MHCRNN models, the single frame is replicated to fill the 5-frame context window. Docstring documents the degenerate-context caveat.

### DIFF
--- a/lightning_pose/api/model.py
+++ b/lightning_pose/api/model.py
@@ -14,7 +14,6 @@ from lightning_pose.data import _IMAGENET_MEAN, _IMAGENET_STD
 from lightning_pose.data.datatypes import MultiviewPredictionResult, PredictionResult
 from lightning_pose.data.utils import convert_bbox_coords
 from lightning_pose.models import ALLOWED_MODELS
-from lightning_pose.models.heatmap_tracker_mhcrnn import HeatmapTrackerMHCRNN
 from lightning_pose.utils import io as io_utils
 from lightning_pose.utils.predictions import generate_labeled_video as generate_labeled_video_fn
 from lightning_pose.utils.predictions import (
@@ -154,9 +153,8 @@ class Model:
 
         Returns:
             {"keypoints": (num_kp, 2) float32 array (x, y) in original frame coords,
-             "confidence": (num_kp,) float32 in [0, 1] -- softmax peak intensity
-              per keypoint, not a calibrated probability.
-              For regression models, confidence is always 1.0.}
+             "confidence": (num_kp,) float32 in [0, 1] -- likelihood/confidence
+              per keypoint. For regression models, confidence is always 1.0.}
 
         Raises:
             ValueError: If frame_rgb has wrong shape/dtype, bbox has non-positive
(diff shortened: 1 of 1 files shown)


lightning_pose/api/model.py
```diff
@@ -116,6 +117,170 @@ def cropped_csv_file_path(self, csv_file_path: str | Path):
...
+        # Normalize: /255, subtract mean, divide std (ImageNet)
+        tensor = resized.astype(np.float32) / 255.0
+        mean = np.array(_IMAGENET_MEAN, dtype=np.float32)
+        std = np.array(_IMAGENET_STD, dtype=np.float32)
+        tensor = (tensor - mean) / std
+
+        # HWC -> CHW, add batch dim
+        tensor = np.transpose(tensor, (2, 0, 1))  # (3, H, W)
+
+        device = next(self.model.parameters()).device
```
themattinthehatt: since the model is a Lightning module I believe you can just call `self.model.device`

lightning_pose/api/model.py
```diff
@@ -116,6 +117,170 @@ def cropped_csv_file_path(self, csv_file_path: str | Path):
...
+        self.model.eval()
+
+        if is_context:
+            # MHCRNN expects (seq_len, 3, H, W) -- replicate frame to 5-frame sequence.
+            # Shape is (5, 3, H, W) not (1, 5, 3, H, W) so it enters
+            # get_representations() via the 4D branch (DALI unlabeled-sequence path).
+            # With all-identical frames, boundary padding is indistinguishable from
+            # non-padding, so get_context_from_sequence produces a valid context window.
+            tensor_t = torch.from_numpy(tensor).unsqueeze(0).expand(5, -1, -1, -1)
+            tensor_t = tensor_t.to(device)
```
themattinthehatt: update this assuming the user has passed in the correct data tensor (i.e. already shape (5, H, W, 3))

lightning_pose/api/model.py
```diff
@@ -116,6 +117,170 @@ def cropped_csv_file_path(self, csv_file_path: str | Path):
...
+        # When bbox is provided, use actual_w/actual_h (not requested bw/bh)
+        # because numpy clips silently when bbox extends beyond frame edges.
+        if bbox is not None:
+            kp[:, 0] = kp[:, 0] * actual_w + bx
+            kp[:, 1] = kp[:, 1] * actual_h + by
+        else:
+            kp[:, 0] *= frame_rgb.shape[1]  # width
+            kp[:, 1] *= frame_rgb.shape[0]  # height
+
+        return {"keypoints": kp, "confidence": conf}
```
themattinthehatt: lines 246-282: it would be better to package the image data into a `HeatmapLabeledBatchDict` and then call `self.model.get_loss_inputs_labeled()`: https://github.com/paninski-lab/lightning-pose/blob/71910dec7d9dce308dcf02e815b227a33dd0d6f6/lightning_pose/models/heatmap_tracker.py#L134 

This will keep you from needing to re-write heatmap/bbox logic, and make things more maintainable/modular. I believe this will allow you to use this function for the [regression tracker](https://github.com/paninski-lab/lightning-pose/blob/71910dec7d9dce308dcf02e815b227a33dd0d6f6/lightning_pose/models/regression_tracker.py#L22) since you will no longer need to call the `run_subpixelmaxima` function directly.

themattinthehatt (approved):

PR #401, opened 2026-05-14
discover calibration file from image path

ksikka:


lightning_pose/data/datasets.py
ksikka, line 978: @themattinthehatt what is this doing? I don't see any mention of imgaug_3d in this file - do we want to skip this when imgaug_3d is false, but camera params are present in the data directory?  (If so i'll need to make a change)

themattinthehatt: yeah I need to make this more transparent...if `imgaug_3d: true` then the `self.imgaug_transform` is constructed without resizing, so that the `if` statement is true

ksikka:
This should be ready to go. Waiting to see test status and review docs in readthedocs live preview.

themattinthehatt (approved):

PR #418, opened 2026-05-31
Fix: synchronize multiview DALI video readers under random_shuffle

LennyAharon:
## Summary

In a multiview pipeline, each view is read by its own `fn.readers.video` operator. DALI auto-assigns a **different seed per operator**, so under `random_shuffle=True` (training) the per-view readers shuffled to unrelated sessions/timepoints. The views in a batch therefore came from different frames, silently breaking the multiview correspondence that the unsupervised losses (e.g. `pca_multiview`) depend on. Training/validation curves looked off because the geometric consistency the losses assume never held.

### Changes
- **`video_pipe`**: new explicit `reader_seed` argument, passed as `seed=` to every per-view `fn.readers.video` so all views shuffle to the same sequence.
- **`PrepareDALI._setup_pipe_dict`**: computes a single `pipeline_seed` (`seed + device_id`) once and uses it for both `seed` and `reader_seed` across all train/predict × base/context pipes (also removes prior duplication).
- **`PrepareDALI.__init__`**: validates multiview inputs — all views must have the same number of sessions and matching per-session frame counts — and raises `ValueError` instead of silently desyncing.
- **`get_videos_in_dir`**: matches view names as non-alphanumeric-delimited tokens (so filenames with infixes like `.short` resolve correctly) and raises if a view yields zero videos.

## Test plan
- [x] Added `test_prepare_dali_multiview_synchronized` (in `tests/data/test_dali.py`): builds the real training loader (3 views, shuffle on), asserts every view reads identical frames across 4 consecutive batches, asserts frames are non-degenerate, and checks both new `ValueError` guards.
- [x] **Verified the test fails without the fix**: temporarily giving each reader a differing seed makes the `np.allclose(frames[:, 0], frames[:, view])` assertion fail — confirming it is a genuine regression guard.
- [x] Test passes with the fix on an L4 GPU (DALI 1.46): `pytest tests/data/test_dali.py::test_prepare_dali_multiview_synchronized`.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/lightning_pose/data/dali.py
+++ b/lightning_pose/data/dali.py
@@ -42,6 +42,7 @@ def video_pipe(
     pad_last_batch: bool = False,
     imgaug: str = "default",
     skip_vfr_check: bool = True,
+    reader_seed: int = 123456,
     # arguments consumed by decorator:
     # batch_size,
     # num_threads,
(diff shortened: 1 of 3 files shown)


themattinthehatt (approved):
looks great 👌

Issue #263, opened 2025-03-11
Feature request: distinguish between occluded and unlabeled keypoints in ground truth data


themattinthehatt:
It has become increasingly important to allow pose estimation models to train on heterogeneous datasets, which are composed of multiple sub-datasets that may be similar in appearance (e.g., head-fixed mice, or [top-down views of mice](https://www.nature.com/articles/s41467-024-48792-2)) but have different subsets of labeled keypoints.

In order to accommodate training on such heterogeneous datasets, Lightning Pose must be able to distinguish between labels missing due to occlusions, and labels missing due to not being part of the original keypoint set for that dataset. Therefore, we propose to introduce a new (optional) coding scheme in the labeled data csv file to account for these different scenarios. We will follow the [COCO format](https://github.com/jin-s13/COCO-WholeBody/blob/master/data_format.md), such that we add a third column `v` for each keypoint which codes the visibility: 

> visibility flag v defined as v=0: not labeled; v=1: labeled but not visible; and v=2: labeled and visible. A keypoint is considered visible if it falls inside the object segment.

By default, keypoints where v=0 are not included in the model loss; keypoints where v=1 behavior will be controlled by the existing config option `training.uniform_heatmap_for_nan_keypoints`; and keypoints where v=2 will have the standard heatmap loss applied.

themattinthehatt:
closed by #440

PR #440, opened 2026-06-11
add visibility column to labels csv

themattinthehatt:
Add per-keypoint visibility flags to label CSV format, and refactor the CSV parsing and heatmap generation pipeline to propagate visibility cleanly.

## Visibility column

Label CSVs now support an optional visible column after each x, y pair (x, y, visible, x, y, visible, …), following the COCO keypoint convention:

* 0: not labeled -- all-zeros (excluded from loss)
* 1: occluded -- uniform (encourages low-confidence output)
* 2: visible -- Gaussian (standard supervised target)

  Detection is inline in parse_label_csv; only the [0, 1, 2] (DLC) header format is supported.

## Refactors

utils/io.py — consolidated CSV parsing into parse_label_csv. Removed has_visibility_column (fully superseded by inline detection). get_keypoint_names is retained; it operates on predictions CSVs (with likelihood column) and has a config fallback, so it serves a different purpose.

data/datatypes.py — BaseLabeledExampleDict now always carries a visibility key: a (K,) int64 tensor when the CSV has a visible column, or torch.zeros(0, dtype=torch.long) (empty-tensor sentinel) when it does not. The sentinel collates cleanly through PyTorch's default_collate.

data/datasets.py — HeatmapDataset.__init__ synthesizes visibility when the CSV has no visible column: NaN keypoints → vis=1 (uniform) or vis=0 (zero) based on the uniform_heatmaps_for_nan_keypoints config flag; valid keypoints → vis=2. This means self.visibility is always populated after init. compute_heatmap now reads visibility directly from example_dict["visibility"] rather than indexing self.visibility by sample index.

data/utils.py — removed uniform_heatmaps parameter from generate_heatmaps; visibility is now encoded permanently in the dataset tensor and passed via example_dict. The visibility=None path (used by loss functions that never pass NaN-coordinate predictions) is unchanged.

losses/losses.py — removed uniform_heatmaps parameter from UnimodalLoss and ReprojectionHeatmapLoss (was already always False at call sites).

## Tests

37 new or updated tests across tests/utils/test_io.py, tests/data/test_utils.py, tests/data/test_datasets.py, and tests/losses/test_losses.py. Coverage includes all four visibility modes (None, vis=0/1/2) for OOB, NaN, and valid keypoints.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -261,24 +261,34 @@ Label CSV files support an optional `visible` column after each `x, y` pair
 | 1 | occluded | uniform (encourages low-confidence output) |
 | 2 | visible | Gaussian (standard supervised target) |
 
-**Detection**: `io_utils.has_visibility_column(csv_file, header_rows)` returns `True` when
-the `coords` header row contains a `visible` entry. Only the `[0, 1, 2]` (DLC) header format
-is supported; all other formats return `False`.
+**Detection**: inline in `parse_label_csv` — the `coords` header row is checked for a `visible`
+entry. Only the `[0, 1, 2]` (DLC) header format is supported; all other formats are treated as
+having no visibility column.
 
 **Parsing** (`BaseTrackingDataset.__init__`): when detected, the raw CSV is reshaped to
 `(N, K, 3)` and split into `self.keypoints (N, K, 2)` (x, y) and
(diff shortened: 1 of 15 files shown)


themattinthehatt:
closes #263

Issue #448, opened 2026-06-24
Multiview transformer broken with transformers >= 5.9


themattinthehatt:
`HeatmapTrackerMultiviewTransformer` crashes during forward pass with `transformers >= 5.9.0`:

```
  IndexError: Dimension out of range (expected to be in range of [-2, 1], but got 2)

  at heatmap_tracker_multiview.py line 204:

  hidden_states = layer(hidden_states)[0]
```

**Cause**

  In transformers < 5.9, ViTLayer.forward returned a tuple, so [0] correctly extracted the hidden states tensor.
  Starting in 5.9.0, ViTLayer.forward returns a plain torch.Tensor. The [0] now indexes dimension 0 of the tensor
  itself, selecting the first batch element and collapsing the shape from (batch, seq, hidden) to (seq, hidden). On
  the next layer iteration, the attention module calls .transpose(1, 2) on the now-2D tensor and raises the
  IndexError.

  The error only surfaces on the second layer iteration (the first layer still produces output, just with wrong
  shape), which is why the traceback points to the attention internals rather than our code directly.

**Fix**

  Check the return type to handle both old and new transformers versions:

  output = layer(hidden_states)
  hidden_states = output[0] if isinstance(output, tuple) else output

 **Affected versions**

  - Works: transformers < 5.9
  - Broken: transformers >= 5.9 (confirmed on 5.9.0 and 5.11.0)

themattinthehatt:
closed by #449

PR #449, opened 2026-06-24
 fix: handle ViTLayer tensor return in transformers >= 5.9

Tejakshaya:
Closes issue #448 

 - Fix `IndexError: dimension out of range` in `HeatmapTrackerMultiviewTransformer.forward_vit` when using
  `transformers >= 5.9`
  - Added `isinstance` check to handle both tuple (old) and tensor (new) return types from `ViTLayer.forward()`

What was wrong:

In `transformers >= 5.9`, `ViTLayer.forward()` returns a plain `torch.Tensor` instead of a tuple and the existing code 
`hidden_states = layer(hidden_states)[0]` indexes dimension 0 of the tensor itself selecting only the first batch element and collapsing the shape from `(batch, seq, hidden)` to `(seq, hidden)` and on the next layer iteration the attention module calls `.transpose(1, 2)` on the now 2D tensor and raises:

IndexError: Dimension out of range (expected to be in range of [-2, 1], but got 2)

the fix <-  This correctly extracts the hidden states regardless of whether the layer returns a tuple (transformers < 5.9) or a tensor (transformers >= 5.9): 

  tensor (transformers >= 5.9):
  ```python
  output = layer(hidden_states)
  hidden_states = output[0] if isinstance(output, tuple) else output

### DIFF
--- a/lightning_pose/models/heatmap_tracker_multiview.py
+++ b/lightning_pose/models/heatmap_tracker_multiview.py
@@ -201,7 +201,8 @@ class HeatmapTrackerMultiviewTransformer(BaseSupervisedTracker):
         else:
             hidden_states = embedding_output
             for layer in vit.layers:
-                hidden_states = layer(hidden_states)[0]
+                output = layer(hidden_states)
+                hidden_states = output[0] if isinstance(output, tuple) else output
             sequence_output = hidden_states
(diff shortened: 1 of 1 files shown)


themattinthehatt:
thanks for the contribution @Tejakshaya! I need to fix some permissions for the CI then will merge this

PR #451, opened 2026-06-24
fix: filter out unlabelled keypoints in SLEAP converter

### LINKED ISSUE #385: SLEAP converter copies unlabelled key points
The SLEAP converter script copies the X, Y coordinates of unlabelled key points to the newly created Lightning Pose project

# Steps to reproduce:

1. Run the SLEAP labelling GUI
2. Create a new project (File->New Project)
3. Add a video to the project (Videos->Add Videos)
4. Add a node to the project (Skeleton->New Node)
5. Add an instance to the project (Instances->New Instance)
6. **Critical step:** Do not click on any key points
7. Save the project (File->Save)
8. Export the labels (Predict->Export Labels Package...->Labeled frames)
9. In a terminal/command prompt, navigate to the lighting-pose scripts folder
10. Run `python sleap2lp.py --slp_file=<path where you exported the labels in step 8> --lp_dir=<some folder>`
11. Open the `CollectedData.csv` file in the folder specified in step 10

# Expected behaviour:

The CSV file has no X, Y coordinates in it

# Actual behaviour:

The CSV file has X, Y coordinates in it

# Additional information

When creating a labelling instance in SLEAP, you do not have to label all points, and you can mark some points as occluded. Right now the `sleap2lp.py` converter script copies the X, Y coordinates of all points, including unlabelled ones. In SLEAP, when you add key points to a frame, it gives arbitrary X, Y coordinates to all key points specified in your labelling config, but doesn't use these for training until you click on them and move them to the correct position. These initially assigned coordinates are essentially random and vanishingly unlikely to be correct, hence should not be exported (or replaced with NaN, or some other way to indicate that those points have not really been labelled).

The `points` dataset in a SLEAP labels package files has keys `"complete"` and `"visible"` containing booleans specifying if a point has been labelled and is visible (at least I assume that's what they mean based on the names, I would need to do more thorough testing to check). Thus the X, Y coordinates for a point are only valid (and therefore should only be exported) if the point's `"complete"` and `"visible"` values are both `True`.

Inserting the following code before the colon on line 115 of `sleap2lp.py` appears to fix the issue, but I have not fully tested it:

` or not kp['visible'] or not kp['complete']`

# Environment information

OS (SLEAP machine): Windows x64
OS (lightning-pose machine): Ubuntu 22.04
Python version (SLEAP environment): 3.14.3
Python version (lightning-pose environment): 3.10.16
SLEAP version: 1.5.1
lightning-pose version: 2.0.5

Please let me know if you need any more info. I am aware my lightning-pose installation is slightly out of date, but looking at the lastest version of `sleap2lp.py` on GitHub it appears that this behaviour is still present in the latest version.

Tejakshaya:
- Filter out unlabelled keypoints in `sleap2lp.py` by checking the `visible` and `complete` flags from the SLEAP
  points dataset
  - Prevents arbitrary/random coordinates from being exported as valid labels

What was wrong: 

When creating keypoints in SLEAP arbitrary X/Y coordinates are assigned to all keypoints in the skeleton even those the user hasn't clicked on or positioned. The `sleap2lp.py` converter was only checking for `NaN` values before exporting coordinates but SLEAP's unlabelled points have real (arbitrary) coordinates - they just have `visible=False` and/or `complete=False so this meant that running the converter on a project with partially labelled frames would export incorrect, essentially random coordinates as valid training data which silently degrades the model performance

Fix:

 Added checks for `kp['visible']` and `kp['complete']` alongside the existing `np.isnan()` check, points that aren't both visible and complete are now exported as `None` (which becomes NaN in the CSV) which ends up matching Lightning Pose's convention for unlabelled keypoints

Closes #385

### DIFF
--- a/scripts/converters/sleap2lp.py
+++ b/scripts/converters/sleap2lp.py
@@ -112,7 +112,11 @@ def extract_labels_from_pkg_slp(file_path: str, base_output_dir: str | None) -> 
                         keypoints_flat = []
                         for kp in points:
                             x, y = kp['x'], kp['y']
-                            if np.isnan(x) or np.isnan(y):
+                            if (
+                                np.isnan(x) or np.isnan(y)
+                                or not kp['visible']
+                                or not kp['complete']
+                            ):
                                 x, y = None, None
                             keypoints_flat.extend([x, y])
 


themattinthehatt:
Thanks again @Tejakshaya!

PR #456, opened 2026-07-10
FIX: Docstring arg names typo in `BaseTrackingDataset`

dherrera1911:
Some argument names in the `BaseTrackingDataset()` docstring didn't match the actual function signature.


themattinthehatt:
good catch @dherrera1911 :pray:

PR #458, opened 2026-07-14
Add inference precision support to CLI and Model API

jjaw123:
Adds the ability to run inference at reduced precision (FP16/BF16), in addition to the
default FP32.

## CLI
`litpose predict <model_dir> <input_path> --precision <fp32|fp16|bf16>` (default: fp32)

## Model API
`Model.from_dir(model_dir, precision="16-mixed")` — precision is set once when the model
is loaded rather than per prediction call, since inference in this codebase runs through
three different paths (two separate `pl.Trainer` instances in `predict_dataset`/
`predict_video`, plus a raw `torch.autocast` forward pass in `predict_frame`). Threading
precision through as a kwarg on every prediction method instead would have meant touching
five method signatures and didn't seem to match the "one model, one way of running it"
model here, so I set it once on `Model` and read it from all three internal call sites.

## Docs
New page at `docs/source/user_guide_advanced/mixed_precision.rst` with accuracy and
forward-pass speed results:
- Training precision affects accuracy slightly (FP16 train lowest error, FP32 train
  highest, in our mirror-mouse-fused experiments).
- Inference precision alone (model trained at FP32, evaluated at FP16/BF16) has no
  measurable effect on accuracy across mirror-mouse-fused, mirror-fish, and crim13.
- Forward-pass speedup depends heavily on batch size: actually *slower* than FP32 at
  batch size 1 (autocast overhead), ~3x for ResNet50 and ~4.4x for ViT-S at batch 32+.

## Tests
Fixed two pre-existing test fixtures (`tests/cli/test_predict.py`,
`tests/utils/test_predictions.py`) that construct `Namespace`/mock objects manually and
were missing the new `precision` field.

Note: I hit a reproducible hang in `tests/models/` in my environment (not `test_base.py`
specifically — it stalls at different points on repeated runs) that appears unrelated to
this change, since none of these commits touch `lightning_pose/models/`. Ran everything
outside that directory clean (576 passed). Flagging in case it's a known issue.

### DIFF
--- a/docs/source/user_guide_advanced/mixed_precision.rst
+++ b/docs/source/user_guide_advanced/mixed_precision.rst
@@ -1,0 +1,151 @@
+.. _mixed_precision:
+
+#####################################
+Mixed Precision Training & Inference
+#####################################
+
+Lightning Pose supports running inference at reduced numerical precision (FP16 or BF16
+"mixed precision") in addition to the default FP32. This can speed up the model's forward
+pass on compatible GPUs. Model weights are always stored and loaded as FP32 on disk --
+precision only affects how the forward pass is computed at inference time, not the checkpoint
+itself.
+
+This page will grow over time as we benchmark more architectures, datasets, and hardware.
+
+Usage
+=====
+
+From the CLI, pass ``--precision`` to ``litpose predict``:
+
+.. code-block:: bash
+
+    litpose predict <model_dir> <input_path> --precision fp16
+
+Valid choices are ``fp32`` (default), ``fp16``, and ``bf16``.
+
+From the Python API, pass ``precision`` to ``Model.from_dir``:
+
+.. code-block:: python
+
+    from lightning_pose.api import Model
+
+    model = Model.from_dir("outputs/2024-01-01/12-00-00", precision="16-mixed")
+
+The Python API uses PyTorch Lightning's native precision strings directly:
+``"32-true"``, ``"16-mixed"``, or ``"bf16-mixed"``.
+
+Results: accuracy
+==================
+
+Two separate questions matter here: does the precision used *during training* affect
+accuracy, and does the precision used *during inference* (independent of training precision)
+affect accuracy?
+
+**Training precision.** Training a model at reduced precision does change accuracy slightly,
+on ``mirror-mouse-fused`` (resnet50_animal_ap10k, 100 train frames, mean over 3 seeds):
+
+.. list-table:: Effect of training precision (same precision used for eval)
+   :widths: 40 30
+   :header-rows: 1
+
+   * - Configuration
+     - Mean pixel error
+   * - FP32 train / FP32 eval
+     - 7.11 px
+   * - FP16 train / FP16 eval
+     - 6.40 px
+   * - BF16 train / BF16 eval
+     - 6.84 px
+
+FP16 training gave the lowest error in this experiment, BF16 in the middle, and FP32 the
+highest -- though the gap is small relative to seed-to-seed variation, so this should not be
+read as "FP16 training is definitively better."
+
+**Inference precision.** Independent of training precision, changing the precision used
+*only at inference time* (model trained at FP32, evaluated at FP16/BF16) has essentially no
+effect on accuracy, across every dataset tested:
+
+.. list-table:: Effect of inference precision (model trained at FP32)
+   :widths: 25 20 20 20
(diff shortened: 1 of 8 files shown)


docs/source/user_guide_advanced/mixed_precision.rst
```diff
@@ -0,0 +1,151 @@
...
+   :header-rows: 1
+
+   * - Configuration
+     - Mean pixel error
+   * - FP32 train / FP32 eval
+     - 7.11 px
+   * - FP16 train / FP16 eval
+     - 6.40 px
+   * - BF16 train / BF16 eval
+     - 6.84 px
```
themattinthehatt: add standard error of the mean to each of these, so we see "x.xx ± y.yy px" where "y.yy" is (the standard deviation of the 3 means, one from each seed), divided by sqrt(3).

should include facemap, crim13, mirror-fish, and marmoset3k results to the table as well for completeness

for each dataset in the table, wrap its name in a link to the huggingface dataset, e.g. https://huggingface.co/datasets/paninski-lab/mirror-mouse-fused

docs/source/user_guide_advanced/mixed_precision.rst
```diff
@@ -0,0 +1,151 @@
...
+     - 7.11 px
+     - 7.11 px
+   * - mirror-fish
+     - 11.52 px
+     - 11.52 px
+     - 11.51 px
+   * - crim13
+     - 17.52 px
+     - 17.52 px
+     - 17.51 px
```
themattinthehatt: same as above - add error bars over seeds, and wrap dataset names in links to HF page

docs/source/user_guide_advanced/mixed_precision.rst
```diff
@@ -0,0 +1,151 @@
...
+
+Deltas are well under 0.01px in every case -- far below seed-to-seed variation. In short:
+**it is safe to run inference at reduced precision regardless of what precision the model
+was trained at.**
```
themattinthehatt: state this conclusion at the very top of the page as well - a reader quickly zipping through this page should immediately see (1) what is the effect on accuracy; (2) what is the effect on speed. If they are interested in the details they can then read further.

docs/source/user_guide_advanced/mixed_precision.rst
```diff
@@ -0,0 +1,151 @@
+.. _mixed_precision:
+
+#####################################
+Mixed Precision Training & Inference
```
themattinthehatt, line 4: great doc overall, thanks for adding this!

lightning_pose/cli/commands/predict.py
```diff
@@ -15,6 +15,14 @@
...
 if TYPE_CHECKING:
     from lightning_pose.api import Model
+    from lightning_pose.api.model import _Precision
+
+# Friendly CLI names -> PyTorch Lightning precision strings used internally.
+_PRECISION_CHOICES: dict[str, _Precision] = {
+    "fp32": "32-true",
+    "fp16": "16-mixed",
+    "bf16": "bf16-mixed",
+}
```
themattinthehatt: to make things easy for the user, use a single set of names for the precision types. we should use "fp32", "fp16", and "bf16" as args for _both_ the CLI and API. so these strings should pass straight from CLI to the API, without going through `_PRECISION_CHOICES`, and then the Model API can map the user-facing strings to the pytorch strings ("32-true, etc.) internally. [Make sure to update this in the doc page as well]

tests/api/test_model.py
```diff
@@ -193,6 +195,19 @@ def test_predict_frame_basic(self, tmp_path, request):
         assert np.all(kp[:, 0] <= 256 + 1)
         assert np.all(kp[:, 1] <= 256 + 1)
 
+    def test_predict_frame_fp16_precision(self, tmp_path, request):
```
themattinthehatt, line 199: this tests that the branch runs, but it doesn't actually test if torch.autocast is working. is there a clearer way to test that functionality exactly?

themattinthehatt (changes requested):
Looks good! Three main comments:

1. we should unify the strings that users provide across the CLI+API; see my comments
2. add results from more datasets to the new doc page
3. make sure failing linting test passes

themattinthehatt (changes requested):
looks good! just need to fix the linting error

themattinthehatt (approved):

PR #466, opened 2026-08-01
Add mixed precision training (AMP) for ~2x speedup

Tejakshaya:
This PR enables mixed precision training (AMP) by adding `precision="16-mixed"` to the `Trainer` in `lightning_pose/train.py`

Mixed precision uses FP16 for most operations which speeds up training and reduces GPU memory usage with no loss in model accuracy.

I benchmarked the performance gain using a standalone ResNet‑50 model (which is teh same backbone thats used in Lightning Pose) on a Google Colab T4 GPU since I had no access to NVIDIA GPUs. 

The test script:
- Ran 20 forward/backward passes with batch size 8 in FP32
- Ran the same passes with AMP (`torch.amp.autocast` + `GradScaler`)
- Measured wall clock time (with `torch.cuda.synchronize()`) and peak GPU memory

Results:
image

On older GPUs or CPU, Lightning will fall back to FP32 with no errors or crashes
precision="16-mixed" is a standard Lightning flag; it works on all GPUs with compute capability ≥ 7.0 (Volta, Turing, Ampere, Ada)

### DIFF
--- a/lightning_pose/train.py
+++ b/lightning_pose/train.py
@@ -419,6 +419,7 @@ def _train(cfg: DictConfig | ListConfig, status_file: Path | None = None) -> Mod
         val_check_interval=val_check_interval,
         log_every_n_steps=cfg.training.log_every_n_steps,
         callbacks=callbacks,
+	precision="16-mixed",
         logger=logger_tb,
         # To understand why we set this, see 'max_size_cycle' in UnlabeledDataModule.
         limit_train_batches=cfg.training.get("limit_train_batches") or steps_per_epoch,


themattinthehatt:
@Tejakshaya thanks for the PR! we actually just merged a PR that does this exact same thing - see #458 

You can also see more in-depth documentation here: https://lightning-pose.readthedocs.io/en/latest/source/user_guide_advanced/mixed_precision.html

Tejakshaya:
Oh! I am SO sorry, I should've checked beforehand.

themattinthehatt:
no problem, we haven't advertised that feature widely yet. glad you'll get some use out of it though!

themattinthehatt:
@Tejakshaya we started playing around with some other ways to speed up LP inference, you can see results here: https://lightning-pose.readthedocs.io/en/latest/source/user_guide_advanced/increasing_inference_speed.html.
In particular, combining fp16 with `torch.compile()` is super simple and speeds things up even more!

We're working now on buidling these options directly in the API/CLI, but I figured you might be interested in the more manual approaches now in the docs. I'll keep you updated.

PR #469, opened 2026-08-03
Add torch.compile() support to Model API and predict CLI

### LINKED ISSUE #467: Add `torch.compile()` support to speed up inference
## Background
 
Benchmarking showed `torch.compile()` gives solid inference speedups with a single-line change — up to ~6x on some model/GPU combinations (see `docs/source/increasing_inference_speed.rst`). We want to expose this through both the Python API and the CLI so users can opt into it without writing any PyTorch code themselves.
 
### 1. Add a `compile()` method to `Model` API
 
In `lightning_pose/api/model.py`, add a method that compiles the model's forward pass in place:
 
```python
def compile(self) -> None:
    """Compile the model's forward pass with torch.compile() for faster inference.
 
    Must be called before running prediction. Compilation is triggered lazily on
    the first inference call (can take tens of seconds); subsequent calls with the
    same input shape reuse the compiled graph. Changing batch size or input
    resolution triggers a new compilation automatically.
    """
    self.model.forward = torch.compile(self.model.forward)
```
 
**Important implementation detail:** compile `self.model.forward` directly and reassign it, rather than wrapping the whole module (`self.model = torch.compile(self.model)`). Some of our prediction code paths call `get_loss_inputs_labeled` rather than going through `__call__`/`forward`, and compiling the whole module silently falls back to the *original* uncompiled submodule for any method other than `forward`/`__call__`. Compiling `forward` directly avoids this trap. (See the reference doc linked above for more detail on why this matters.)
 
### 2. Usage pattern — Python API
 
```python
 
model = Model.from_dir("path/to/model_dir")
model.compile()
result = model.predict_on_video_file("path/to/video.mp4")
```
 
### 3. Add a `--compile` flag to the CLI's `predict` command
 
```bash
litpose predict /path/to/model_dir /path/to/video.mp4 --compile
```
 
Wire this up where `litpose predict` constructs its `Model` and calls `predict_on_video_file`/equivalent — after loading the model, call `.compile()` if the flag is set:
 
```python
parser.add_argument(
    "--compile",
    action="store_true",
    default=False,
    help="Compile the model with torch.compile() for faster inference.",
)
 
# ...
model = Model.from_dir(args.model_dir)
if args.compile:
    model.compile()
result = model.predict_on_video_file(args.video_path)
```
 
## Testing
 
- Add a unit test that calls `model.compile()` and runs a prediction on a labeled csv, asserting it completes without error and returns predictions of the expected shape. Add another unit tests that runs a prediction on the short test video fixture. Both unit tests can just be for a singleview model.
- Add an accuracy test: compare predictions from a compiled model against the same eager model on a few real frames (label csv file), and assert max pixel deviation is small (e.g. under 0.1px) — consistent with the numbers already in the benchmarking doc. This guards against a future PyTorch version changing compile behavior in a way that actually affects outputs. This test only needs to run on a singleview model.
- Test that calling `.compile()` twice, or compiling then running on a different batch size, doesn't error (should just trigger recompilation).
- Add a CLI test that `--compile` runs end-to-end on a small test video.

## Documentation
 
- Update the existing inference speed doc, showing the `model.compile()` and `litpose predict --compile` usage patterns above.
- Note that the first call after compiling will be slower (compilation overhead) and this is expected.

This issue is scoped to `torch.compile()` only. Please don't add ONNX/TensorRT export or loading — that will be handled as a separate, follow-up piece of work.

jjaw123:
Closes #467.

Adds `torch.compile()` support through both the Python API and the CLI.

## API

```python
model = Model.from_dir("path/to/model_dir")
model.compile()
result = model.predict_on_video_file("path/to/video.mp4")
```

`compile()` wraps `self.model.forward` and reassigns it, rather than wrapping the whole module, for the reason described in the issue: `predict_frame` and friends call `get_loss_inputs_labeled`, and a compiled module silently falls back to the uncompiled submodule for anything that isn't `forward`/`__call__`.

**Deviation from the issue snippet:** `Model.model` is `None` until the first prediction (weights load lazily in `_load()`), so `self.model.forward = ...` would raise `AttributeError` when called right after `from_dir` — which is the usage the issue shows. `compile()` therefore calls `self._load()` first. This means `compile()` eagerly loads weights, trading away lazy loading for errors that surface immediately; happy to invert that (defer via a flag applied inside `_load()`) if you'd rather preserve laziness.

A `_compiled` flag makes repeat calls a no-op instead of double-wrapping `forward`.

## CLI

```
litpose predict /path/to/model_dir /path/to/video.mp4 --compile
```

## Measurements

Singleview ViT-S on mirror-mouse-fused, 789 labeled frames, 1xT4:

| | time | notes |
|---|---|---|
| eager | 51.7s | |
| compiled, 1st pass | 53.1s | includes compilation |
| compiled, steady state | 8.8s | **5.86x** |

Max abs pixel deviation vs eager: **0.0038px** (mean 0.000016px); 0 of 22,092 coordinates exceed 0.1px. Consistent with the numbers already in the benchmarking doc.

Note `Not enough SMs to use max_autotune_gemm` on T4 — Inductor declines that optimization on this GPU, so speedup may differ on A100/L4.

## Tests

`TestCompile` in `tests/api/test_model.py` (GPU-marked, singleview): compile-then-predict on a labeled CSV and on a video, compiled-vs-eager accuracy under 0.1px, idempotency, and a changing-input-shape case that forces recompilation. `test_compile_loads_model` pins the lazy-load behavior above.

CLI: parser tests for the flag and its default, plus `handle()` tests asserting `compile()` is called when set, not called otherwise, and applied on the multiview path too. Added `compile=False` to the `_make_args` fixture.

48 passed on a T4; 22 passed CPU-only.

## Docs

Rewrote the `torch.compile` section of `increasing_inference_speed.rst` to use the supported API and CLI, and kept your "why forward, not the module" note as implementation rationale.

One thing I left alone: the ONNX section says the session is plugged in "the same way as above," which referred to the manual `forward` assignment that section no longer shows. Left it since ONNX is out of scope here — worth a tidy-up whenever the export work lands.

### DIFF
--- a/docs/source/user_guide_advanced/increasing_inference_speed.rst
+++ b/docs/source/user_guide_advanced/increasing_inference_speed.rst
@@ -139,35 +139,40 @@ The tutorials below assume you've already trained a model with ``litpose train``
 torch.compile
 -------------
 
-``torch.compile()`` compiles a callable, so to make sure it's actually used everywhere
-Lightning Pose's prediction methods call into the model (some of which call
-``get_loss_inputs_labeled`` directly rather than the module's ``__call__``), compile the
-``forward`` method itself and assign it back onto the model, rather than wrapping the whole
-module:
+``Model.compile()`` compiles the model's forward pass in place. Call it after loading the
+model and before running prediction:
 
 .. code-block:: python
 
-    import torch
     from lightning_pose.api import Model
 
     model = Model.from_dir("path/to/model_dir")
-    model.model.forward = torch.compile(model.model.forward)
+    model.compile()
 
     # use as normal -- the compiled graph is now used internally
     result = model.predict_on_video_file("path/to/video.mp4")
 
-The first call triggers compilation (can take tens of seconds); subsequent calls with the
-same input shape reuse the compiled graph. Changing batch size or input resolution triggers a
-new compilation automatically.
+The same thing is available from the CLI with the ``--compile`` flag:
+
+.. code-block:: console
+
+    litpose predict /path/to/model_dir /path/to/video.mp4 --compile
+
+The first prediction after compiling triggers compilation and is therefore *slower* than an
+uncompiled run -- expect tens of seconds of one-time overhead. This is expected. Subsequent
+calls with the same input shape reuse the compiled graph. Changing batch size or input
+resolution triggers a new compilation automatically. Calling ``compile()`` more than once is
+a no-op.
 
 .. note::
-   Wrapping the whole module (``model.model = torch.compile(model.model)``) also works if
-   you're calling ``model.model(images)`` directly in your own code. It does **not** reliably
-   engage the compiled graph when going through ``predict_on_video_file`` / ``predict_frame``
-   / ``predict_on_label_csv``, because those call ``get_loss_inputs_labeled`` rather than
-   ``forward`` -- and calling anything other than ``forward``/``__call__`` on a compiled module
-   silently falls back to the *original*, uncompiled submodule. Compiling ``forward`` directly
-   (as above) avoids this trap.
+   ``compile()`` compiles the model's ``forward`` method and assigns it back onto the model,
+   rather than wrapping the whole module. Wrapping the module (``model.model =
+   torch.compile(model.model)``) works if you call ``model.model(images)`` directly in your
+   own code, but does **not** reliably engage the compiled graph when going through
+   ``predict_on_video_file`` / ``predict_frame`` / ``predict_on_label_csv``, because those
+   call ``get_loss_inputs_labeled`` rather than ``forward`` -- and calling anything other
+   than ``forward``/``__call__`` on a compiled module silently falls back to the *original*,
+   uncompiled submodule.
 
 .. _usage_onnx_runtime:
 
--- a/docs/source/user_guide_advanced/increasing_inference_speed.rst
+++ b/docs/source/user_guide_advanced/increasing_inference_speed.rst
@@ -177,8 +177,9 @@ a no-op.
 .. note::
    ``torch.compile()`` uses the TorchInductor backend, which generates Triton kernels and
    therefore requires a GPU of CUDA compute capability **7.0 or higher** (Volta and newer).
-   On older GPUs -- for example the GTX 10-series (capability 6.x) -- the first prediction
-   after calling ``compile()`` raises ``BackendCompilerFailed``. Check your device with
+   On older GPUs -- for example the GTX 10-series (capability 6.x) -- ``compile()`` raises
+   a ``RuntimeError`` immediately, rather than failing partway through the first
+   prediction. Check your device with
    ``python -c "import torch; print(torch.cuda.get_device_capability())"``.
 
 .. _usage_onnx_runtime:
--- a/docs/source/user_guide_advanced/increasing_inference_speed.rst
+++ b/docs/source/user_guide_advanced/increasing_inference_speed.rst
@@ -174,6 +174,13 @@ a no-op.
    than ``forward``/``__call__`` on a compiled module silently falls back to the *original*,
    uncompiled submodule.
 
+.. note::
+   ``torch.compile()`` uses the TorchInductor backend, which generates Triton kernels and
+   therefore requires a GPU of CUDA compute capability **7.0 or higher** (Volta and newer).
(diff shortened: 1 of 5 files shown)


jjaw123:
@themattinthehatt ready for review whenever you get a chance.

One testing deviation worth flagging explicitly: the issue asks for a CLI test that
`--compile` runs end-to-end on a small test video. `tests/cli/test_predict.py` has no
end-to-end harness today — it's parser tests plus mock-based `handle()` tests — so I matched
that style and asserted `model.compile()` is invoked when the flag is set, not invoked
otherwise, and applied on the multiview path too.

The actual end-to-end coverage (compile, then predict on the short video fixture) lives in
`TestCompile::test_compile_then_predict_on_video_file`, which uses the same
`videos/test_vid.mp4` fixture the issue refers to — just driven through the API rather than
the CLI. Happy to add a subprocess-level CLI test if you'd rather have it exercised through
`litpose` itself.

themattinthehatt:
thanks for flagging! your testing approach looks good

jjaw123:
@themattinthehatt 

Pushed a fix for the CI failure. The `test` job died on exactly the four tests that execute a compiled forward pass:

> RuntimeError: Found NVIDIA GeForce GTX 1080 Ti which is too old to be supported by the triton GPU compiler, which is used as the backend. Triton only supports devices of CUDA Capability >= 7.0, but your device is of CUDA capability 6.1

ax05's 1080 Ti is capability 6.1 and TorchInductor's Triton backend needs ≥ 7.0 (Volta and newer), so those four now carry a `skipif` on `torch.cuda.get_device_capability()`, and I added a note about the requirement to the docs page. The other two compile tests still run on CI — compilation is lazy, so they exercise the full body of `compile()` without ever invoking Inductor — which should keep patch coverage intact.

Worth knowing that this means the compiled inference path won't get CI coverage on this runner. I verified it by hand on a T4 instead (5.86x steady-state, 0.0038px max deviation over 789 frames, table above), but that won't catch a future regression the way a test would.

One open question, not blocking: right now someone on a pre-Volta card gets no error from `model.compile()` and then a Dynamo traceback partway through their prediction. Would you want `compile()` to check the device up front and raise something readable (or warn and no-op)? I left it alone since the issue didn't ask for it, but happy to add it if you'd prefer the earlier, clearer failure.

themattinthehatt:
thanks for tracking that down!

yes it would be great if you can check the device up front and raise a readable error

jjaw123:
Done — `compile()` now checks `torch.cuda.get_device_capability()` up front and raises a `RuntimeError` naming the device, its capability, and the 7.0 requirement, before the checkpoint is even loaded. The check keys off the available CUDA device rather than the model's current device, since at `compile()` time the model may still be on CPU and only get moved to the GPU later by the `Trainer` — which is exactly how the original failure surfaced.

The new test mocks the capability instead of skipping, so this path is covered on the CI runner even though the four compiled-inference tests can't run there. Docs note updated to describe the new behavior.

jjaw123:
CI has failed twice on the self-hosted runner, both ending in `client_loop: send disconnect: Broken pipe` / exit 255, but with very different durations — 5m01s and then 2h14m.

One detail that might narrow it down: both failures stop at exactly the same point in the log. The last line is the `git checkout` of `FETCH_HEAD`, and neither ever prints `Collecting package metadata (current_repodata.json)`, which the last successful run emitted immediately afterward. So the remote process seems to stall right as conda starts fetching repodata from anaconda.org, and the broken pipe looks like a consequence of the session going silent rather than the cause. Might be worth checking whether conda's solve/fetch is hanging on ax05 rather than treating it as a network drop on the SSH side.

I retriggered once with an empty commit (I can't re-run `pull_request_target` jobs without base-repo write access) and got the 2h14m hang, so I've stopped rather than burn more runner time. Also can't touch the workflow from a fork PR, since `pull_request_target` runs the base branch's workflow file.

On the code itself: the last CI run that actually executed the suite was green apart from the four torch.compile tests failing on the 1080 Ti, which the `skipif` commit addressed. Since then, verified locally — 49 passed on a T4 (`tests/api/test_model.py` + `tests/cli/test_predict.py`, no skips) and `pyright` clean project-wide.

themattinthehatt:
no worries, it happens sometimes. I can keep retriggering on my end until it works.

EDIT: tests are passing now, I'll review the PR tomorrow!

themattinthehatt (approved):

PR #471, opened 2026-08-05
Add ONNX export and runtime support to Model API and predict CLI

### LINKED ISSUE #470: add ONNX runtime support to speed up inference
# Add ONNX export and inference runtime support

## Background

Benchmarking showed ONNX Runtime gives solid inference speedups over eager PyTorch, and — unlike `torch.compile()` — produces a portable artifact that can be deployed without a full PyTorch install (see `docs/source/increasing_inference_speed.rst`). This issue adds the ability to export a trained model to ONNX and run inference through it, via both the Python API and the CLI.

This is a follow-up to #467 (`torch.compile()` support, merged in #469). TensorRT support will be a separate follow-up issue that builds on top of this one — please don't add TensorRT in this PR.

Please read through `lightning_pose/api/model.py` before starting — several things below (checkpoint resolution, the `precision` attribute, lazy loading) already exist there and should be reused, not reimplemented.

## Directory structure

LP model directories already look like this:

```
model_dir/
├── config.yaml
├── tb_logs/.../checkpoints/epoch=214-step=12685-best.ckpt
├── predictions.csv
├── video_preds/
└── image_preds/
```

This issue adds one new fixed subdirectory:

```
model_dir/
├── config.yaml
├── tb_logs/.../checkpoints/epoch=214-step=12685-best.ckpt
├── exports_onnx/
│   ├── epoch=214-step=12685-best_fp16.onnx
│   └── epoch=214-step=12685-best_fp32.onnx   # only if exported
└── ...
```

Users never pass export paths themselves — `export()` always writes here, and `from_dir()` always reads from here. This is a deliberate choice: exposing path arguments tends to create more bugs and confusion than it solves.

The exported filename is the checkpoint's own filename (stem) with `_{onnx_precision}` appended, rather than a fixed `model_{precision}.onnx`. This keeps the fixed-directory convenience while making it unambiguous which checkpoint an export came from if a model_dir ever has exports sitting next to a different or updated checkpoint.

Follow the pattern already used for `image_preds_dir()`, `video_preds_dir()`, etc. in `Model` — add an `exports_onnx_dir()` method returning `self.model_dir / "exports_onnx"` (a `Path`, matching how `model_dir` is already stored) rather than building the path inline wherever it's needed.

## A naming note before you start: `precision` is already taken

`Model` already has a `precision` attribute and `from_dir(..., precision=...)` argument (`"fp32"|"fp16"|"bf16"`) — it controls the autocast/PyTorch-Lightning-trainer precision used during **eager** inference, and has nothing to do with which exported file gets loaded. Don't reuse that name for ONNX export selection or you'll create two different things called `precision` with different meanings depending on `runtime`.

Use **`onnx_precision`** for the new concept throughout (export-time weight precision baked into the `.onnx` file itself): `model.export("onnx", onnx_precision="fp16")`, `Model.from_dir(path, runtime="onnx", onnx_precision="fp16")`, `--onnx-precision` on the CLI. This also only supports `"fp32"`/`"fp16"` (no `bf16"` — ONNX export doesn't need a third option here), which is a further reason not to conflate it with the existing 3-way `precision`.

### 1. Add an `export()` method to `Model`

```python
def export(self, runtime: str, onnx_precision: str = "fp16") -> Path:
    """Export the model to an optimized inference format.

    Currently supports runtime="onnx". Writes to a fixed location inside
    the model directory (exports_onnx_dir() / "{ckpt_stem}_{onnx_precision}.onnx")
    and returns the path to the exported file.
    """
    if runtime != "onnx":
        raise ValueError(f"Unsupported export runtime: '{runtime}'. Only 'onnx' is currently supported.")
    if onnx_precision not in ("fp32", "fp16"):
        raise ValueError(f"Unsupported onnx_precision: '{onnx_precision}'. Use 'fp32' or 'fp16'.")

    self._load()  # weights are lazy-loaded; export needs a real forward pass to trace
    if self.model is None:
        raise RuntimeError('model failed to load; self.model is None after _load()')

    ckpt_file = io_utils.ckpt_path_from_base_path(
        base_path=str(self.model_dir), model_name=self.cfg.model.model_name
    )
    stem = Path(ckpt_file).stem  # e.g. "epoch=214-step=12685-best"

    export_dir = self.exports_onnx_dir()
    export_dir.mkdir(parents=True, exist_ok=True)
    onnx_path = export_dir / f"{stem}_{onnx_precision}.onnx"

    resize_h = self.cfg.data.image_resize_dims.height
    resize_w = self.cfg.data.image_resize_dims.width
    # NOTE: multiview models use a different dummy input shape,
    # (1, num_views, 3, H, W) instead of (1, 3, H, W). Use
    # self.config.is_multi_view() (see predict_on_video_file_multiview
    # for the existing usage pattern) to branch here.
    dummy = torch.randn(1, 3, resize_h, resize_w, device=self.model.device)

    module_to_export = self.model
    if onnx_precision == "fp16":
        module_to_export = module_to_export.half()
        dummy = dummy.half()

    torch.onnx.export(
        module_to_export,
        dummy,
        str(onnx_path),
        input_names=["images"],
        output_names=["heatmaps"],
        dynamic_axes={"images": {0: "batch"}, "heatmaps": {0: "batch"}},
        opset_version=17,
        do_constant_folding=True,
    )
    return onnx_path
```

**Note on `self._load()`:** `Model.model` is `None` until weights are lazy-loaded (see `_load()` in `model.py`). `export()` needs real weights to trace a forward pass, so it must call `self._load()` up front — same as `compile()` already does.

**Note on checkpoint resolution:** use `io_utils.ckpt_path_from_base_path(base_path=str(self.model_dir), model_name=self.cfg.model.model_name)` — this is the same function `_load()` already uses internally to pick a checkpoint file. `export()` and the loading path below (`_attach_onnx_runtime`) both need to agree on which checkpoint they mean, so both should call this exact function rather than each doing their own resolution.

**Note on fp16 export:** exporting at fp16 means casting the module (and the dummy input) to half precision before tracing, not just an output-dtype relabeling. Double check this doesn't leave any of the model's internal buffers in fp32 in a way that breaks the export — flag in the PR if you hit issues here.

### 2. Extend `Model.from_dir` with a `runtime` argument

```python
@staticmethod
def from_dir(
    model_dir: str | Path,
    precision: _Precision = "fp32",
    runtime: str = "eager",
    onnx_precision: str | None = None,
) -> Model:
    """Create a Model instance for a model stored at model_dir.

    runtime="eager" (default): loads the trained checkpoint as usual;
        `precision` controls autocast precision as it does today.
    runtime="onnx": loads an ONNX Runtime session from exports_onnx_dir(),
        built with model.export("onnx", ...) beforehand. `precision` is
        ignored in this mode -- the exported file's own precision is what
        runs. `onnx_precision` selects *which* exported file to load.
    onnx_precision: only used when runtime="onnx". If omitted and exactly
        one ONNX export exists, it's used automatically. If omitted and
        more than one exists, raises an error asking the user to disambiguate.
    """
    model = Model.from_dir2(model_dir, precision=precision)  # existing construction, unchanged

    if runtime == "eager":
        return model
    elif runtime == "onnx":
        model._attach_onnx_runtime(onnx_precision)
        return model
    else:
        raise ValueError(f"Unsupported runtime: '{runtime}'. Use 'eager' or 'onnx'.")
```

```python
def _attach_onnx_runtime(self, onnx_precision: str | None) -> None:
    self._load()  # same lazy-load consideration as compile() / export()
    if self.model is None:
        raise RuntimeError('model failed to load; self.model is None after _load()')

    ckpt_file = io_utils.ckpt_path_from_base_path(
        base_path=str(self.model_dir), model_name=self.cfg.model.model_name
    )
    stem = Path(ckpt_file).stem  # must match export()'s resolution exactly

    export_dir = self.exports_onnx_dir()
    pattern = f"{stem}_{onnx_precision}.onnx" if onnx_precision is not None else f"{stem}_*.onnx"
    candidates = sorted(export_dir.glob(pattern))

    if len(candidates) == 0:
        raise FileNotFoundError(
            f"No ONNX export found for checkpoint '{stem}'"
            + (f" with onnx_precision='{onnx_precision}'" if onnx_precision else "")
            + f" in {export_dir}. Run model.export('onnx') first."
        )
    if len(candidates) > 1:
        found = [c.name for c in candidates]
        raise ValueError(f"Multiple ONNX exports found: {found}. Pass onnx_precision= to disambiguate.")

    onnx_path = candidates[0]
    session = ort.InferenceSession(
        str(onnx_path), providers=["CUDAExecutionProvider", "CPUExecutionProvider"]
    )

    # Guard against a silent fallback to CPU. onnxruntime does not raise if
    # CUDAExecutionProvider fails to load (e.g. onnxruntime-gpu's CUDA/cuDNN
    # major version doesn't match the installed PyTorch/CUDA) -- it just runs
    # on CPU instead, which "works" but is drastically slower with no
    # indication why. Fail loudly here instead.
    if torch.cuda.is_available() and "CUDAExecutionProvider" not in session.get_providers():
        raise RuntimeError(
            "ONNX Runtime failed to load CUDAExecutionProvider even though a CUDA "
            "device is available -- inference would silently run on CPU instead. "
            "This is usually a CUDA/cuDNN major version mismatch between "
            "onnxruntime-gpu and your installed PyTorch. See the installation "
            "section of docs/source/increasing_inference_speed.rst for how to "
            "check and fix this."
        )

    def onnx_forward(images):
        io_binding = session.io_binding()
        io_binding.bind_input(
            name="images", device_type="cuda", device_id=0,
            element_type=np.float32, shape=tuple(images.shape),
            buffer_ptr=images.contiguous().data_ptr(),
        )
        io_binding.bind_output(name="heatmaps", device_type="cuda", device_id=0)
        session.run_with_iobinding(io_binding)
        return torch.from_numpy(io_binding.copy_outputs_to_cpu()[0]).to(images.device)

    self.model.forward = onnx_forward  # type: ignore[method-assign]  -- see compile() for the same pattern
    self._runtime = "onnx"
```

This mirrors `compile()`: reassign `forward` directly rather than wrapping the whole module, so all the existing prediction code paths (which don't always go through `__call__`, e.g. `get_loss_inputs_labeled`) still hit the new implementation.

**New: track which runtime is active.** Add a `_runtime: str = "eager"` attribute to `Model` (alongside the existing `_compiled` flag), set to `"onnx"` in `_attach_onnx_runtime()`. Then update `compile()` to raise if `self._runtime != "eager"`:

```python
def compile(self) -> None:
    if self._runtime != "eager":
        raise RuntimeError(
            f"compile() is only supported for runtime='eager', but this model was "
            f"loaded with runtime='{self._runtime}'. torch.compile() has no effect "
            f"on an ONNX Runtime session."
        )
    ...  # rest unchanged
```

This guard doesn't exist yet in `compile()` from #469 — it couldn't, since `runtime` didn't exist at the time. Adding it is in scope for this PR.

### 3. Usage pattern — Python API

```python

model = Model.from_dir("path/to/model_dir")
model.export("onnx", onnx_precision="fp16")

onnx_model = Model.from_dir("path/to/model_dir", runtime="onnx", onnx_precision="fp16")
result = onnx_model.predict_on_video_file("path/to/video.mp4")
```

### 4. CLI changes

Add an `export` command:

```bash
litpose export /path/to/model_dir --runtime onnx --onnx-precision fp16
```

Add `--runtime` and `--onnx-precision` flags to `predict`:

```bash
litpose predict /path/to/model_dir /path/to/video.mp4 --runtime onnx --onnx-precision fp16
```

```python
# export command
export_parser.add_argument("--runtime", choices=["onnx"], default="onnx")
export_parser.add_argument("--onnx-precision", choices=["fp32", "fp16"], default="fp16")
# ...
model = Model.from_dir(args.model_dir)
model.export(args.runtime, onnx_precision=args.onnx_precision)

# predict command
predict_parser.add_argument("--runtime", choices=["eager", "onnx"], default="eager")
predict_parser.add_argument("--onnx-precision", choices=["fp32", "fp16"], default=None)
# ...
model = Model.from_dir(
    args.model_dir,
    precision=args.precision,       # existing eager/autocast flag, unrelated to the above
    runtime=args.runtime,
    onnx_precision=args.onnx_precision,
)
```

Note `--compile` and `--runtime onnx` are mutually exclusive — `torch.compile()` has no effect on an ONNX Runtime session. This is enforced by the `compile()` guard added above; make sure the CLI's error message (if it validates args itself before calling `Model.from_dir`) is consistent with that, rather than failing with a confusing raw traceback.

## Testing

- Unit test: `export("onnx")` produces a file at the expected path (checkpoint stem + `onnx_precision`), for both `fp32` and `fp16`.
- Unit test: `from_dir(runtime="onnx")` with no `onnx_precision` picks the file automatically when exactly one exists, and raises a clear error when zero or multiple exist.
- Unit test: `export()`'s checkpoint resolution and `_attach_onnx_runtime()`'s resolution agree — i.e. loading finds exactly what exporting produced, since both call `io_utils.ckpt_path_from_base_path`.
- Accuracy test: compare predictions from the ONNX runtime against the same eager model on a few real frames, asserting max pixel deviation is small (e.g. under 0.1px) — consistent with the numbers already in the benchmarking doc.
- Test that `export()` works correctly whether or not weights were already loaded beforehand (covers the lazy-loading interaction).
- Test that `compile()` raises when called on a model loaded with `runtime="onnx"`.
- Test the `_attach_onnx_runtime()` provider guard: mock `session.get_providers()` to omit `CUDAExecutionProvider` while CUDA is reported available, and assert a `RuntimeError` is raised instead of silently proceeding on CPU.
- CLI tests: following the pattern used for `--compile` in #469, mock `Model.export` / `Model.from_dir` and assert the CLI parses `--runtime`, `--onnx-precision`, etc. correctly and calls through with the right arguments — not an end-to-end run. End-to-end coverage of the actual export/inference behavior belongs in the API-level unit tests above, not duplicated at the CLI layer.
- Test on both a singleview and a multiview model, since the dummy input shape differs.

## Documentation

- Add an **installation** section to the ONNX part of `docs/source/increasing_inference_speed.rst`, before the usage snippet, covering:
  - `pip install onnxruntime-gpu` installs a wheel pinned to a specific CUDA major version at build time (currently CUDA 12.x/13.x by default) -- it does not detect or match whatever CUDA the user already has.
  - How to check the CUDA version their installed PyTorch is using (`python -c "import torch; print(torch.version.cuda)"`) and match `onnxruntime-gpu` to it; link to ONNX Runtime's install-selector page for anyone not on a default CUDA 12.x/13.x setup (e.g. still on CUDA 11.x, which requires a different install command since it's no longer on PyPI by default).
  - Note that `onnxruntime-gpu` can reuse PyTorch's own bundled CUDA/cuDNN libraries (no separate system CUDA toolkit needed) as long as `torch` is imported first -- which it already is here, since `Model` imports `torch` at module level -- so most users on a working PyTorch+CUDA setup shouldn't need to install anything beyond `pip install onnxruntime-gpu` itself.
  - What the new `RuntimeError` from `_attach_onnx_runtime()`'s provider guard (see above) means and how to resolve it, since that's the error users will actually hit if their setup is mismatched.
- Update the ONNX section of `docs/source/increasing_inference_speed.rst` to reflect the new `export()` / `from_dir(runtime="onnx")` API instead of the current manual code snippet, with the installation section above appearing first, before the usage snippet. That section currently says the ONNX session gets plugged in "the same way as above," referring to a manual `forward` assignment snippet that no longer exists after the `torch.compile()` docs rewrite in #469 — this is a good time to clean that up.
- Document the `exports_onnx/` directory convention, the distinction between `precision` (eager autocast) and `onnx_precision` (export-time weight precision), and the `litpose export` / `--runtime onnx` CLI usage.
- Add `exports_onnx/` to the model directory tree documented in `docs/source/directory_structure_reference/model_dir_structure.rst`, alongside the existing entries for `tb_logs/`, `video_preds/`, `image_preds/`, etc.

This issue is scoped to ONNX only. Please don't add TensorRT export or the TensorRT runtime — that's a separate follow-up issue that will build on top of this one (same `.onnx` file, different execution provider). Also don't worry about pinning exact `onnxruntime-gpu` version requirements yet — dependency/installation details will be handled separately.

jjaw123:
Closes #470.

Adds ONNX export and inference-runtime support to the Model API and the CLI. TensorRT is deliberately out of scope.

## What's here

- `Model.export("onnx", onnx_precision=...)` writes to `exports_onnx_dir()` as `{ckpt_stem}_{onnx_precision}.onnx`.
- `Model.from_dir(..., runtime="onnx", onnx_precision=...)` builds an ONNX Runtime session and rebinds `forward`, mirroring `compile()` so paths that call `get_loss_inputs_labeled` directly are covered.
- New `_runtime` attribute; `compile()` raises when `_runtime != "eager"`.
- `litpose export <model_dir> --runtime onnx --onnx-precision fp16`, plus `--runtime` / `--onnx-precision` on `litpose predict`. `--compile` with `--runtime onnx` is rejected in the CLI before the model is built, so the message matches the `compile()` guard rather than surfacing as a traceback.
- Provider guard raising when CUDA is available but `CUDAExecutionProvider` didn't load.
- Docs: `onnxruntime-gpu` installation section, ONNX section rewritten around the new API, `exports_onnx/` added to the model-directory reference. The TensorRT snippet was rewired to define its own variables, since it previously continued from the manual recipe that this PR removes.

## Three things I'd like your call on

**1. Context models.** The issue covers singleview and multiview dummy-input shapes. Context (MHCRNN) models need a third, `(1, T, 3, H, W)`. I added that branch with `T = 5`, matching the convention documented in `predict_frame`. But you may prefer `export()` to refuse context models outright rather than produce an export whose temporal behavior nobody has validated. Easy to switch to a `NotImplementedError`.

**2. `heatmap.py:211` registers `torch.tensor` results as constants in the trace.** Torch warns this "might cause the trace to be incorrect." The accuracy test passes at the config's resize dimensions, so it looks benign there — but that only covers the one resolution we tested. If that constant is shape-derived, an export reused at a different input resolution could be silently wrong. You know that head better than I do.

**3. CPU fallback.** The issue's `onnx_forward` hardcodes `device_type="cuda"`, but the provider guard only fires when `torch.cuda.is_available()`, so a CPU-only machine would build a session fine and then fail on first forward. I added a CPU branch. If ONNX should just be GPU-only, dropping it simplifies the code.

## Two corrections to the sketch

**`.half()` mutates in place.** `module_to_export = module_to_export.half()` returns `self`, so exporting fp16 from a live `Model` would leave `self.model` in half precision for every later eager call on that object. `export()` traces a `deepcopy` instead. Covered by `test_fp16_export_leaves_live_model_in_fp32`.

**`element_type=np.float32` is hardcoded.** An fp16 export expects fp16 input buffers. Input dtype is now read from the session's input metadata.

Two smaller ones: checkpoint resolution moved into a shared `_ckpt_stem()` with a `None` guard, so an untrained model directory raises the same `FileNotFoundError` as `_load()` rather than a `TypeError` from `Path(None)`; and `import onnxruntime` is deferred into `_attach_onnx_runtime()` so it stays optional for eager users.

## Testing

`tests/cli/`: 120 passed. `tests/api/test_model.py -k "Export or OnnxRuntime or Compile"`: 24 passed on a Tesla T4 (torch 2.8.0+cu128, onnxruntime 1.28.0, `CUDAExecutionProvider` confirmed active).

Covers export paths for both precisions, auto-selection with zero/one/multiple exports, export-vs-load checkpoint agreement, the lazy-load interaction, `compile()` rejection under `runtime="onnx"`, a mocked provider-guard test, and singleview + multiview. `test_onnx_matches_eager_predictions` confirms max pixel deviation under 0.1px against the eager model on real frames.

## Follow-up, not in this PR

Torch 2.9 makes the `dynamo=True` exporter the default and the current TorchScript path is deprecated. Works on 2.8, but worth its own issue.

### DIFF
--- a/docs/source/user_guide_advanced/increasing_inference_speed.rst
+++ b/docs/source/user_guide_advanced/increasing_inference_speed.rst
@@ -203,6 +203,16 @@ you already have -- so check what your PyTorch install is using first:
 Then use that CUDA version to select the proper installation option from ONNX Runtime's
 `installation selector <https://onnxruntime.ai/getting-started>`_.
 
+Exporting additionally requires the ``onnx`` package, which ``onnxruntime-gpu`` does not pull
+in as a dependency:
+
+.. code-block:: console
+
+    pip install onnx
+
+Only ``Model.export()`` needs it. A machine that just runs an already-exported ``.onnx`` file
+needs ``onnxruntime-gpu`` alone.
+
 Usage
 ~~~~~
 
--- a/docs/source/user_guide_advanced/increasing_inference_speed.rst
+++ b/docs/source/user_guide_advanced/increasing_inference_speed.rst
@@ -187,41 +187,21 @@ a no-op.
 ONNX Runtime
 ------------
 
+.. _onnx_installation:
+
 Installation
 ~~~~~~~~~~~~
 
-ONNX Runtime is an optional dependency:
-
-.. code-block:: console
-
-    pip install onnxruntime-gpu
-
-``onnxruntime-gpu`` wheels are built against a specific CUDA major version at build time --
-currently CUDA 12.x/13.x for the default PyPI package. The wheel does **not** detect or adapt
-to whatever CUDA you already have, so check what your PyTorch install is using and match it:
+ONNX Runtime is an optional dependency. ``onnxruntime-gpu`` wheels are built against a specific
+CUDA major version at build time, and the wheel does **not** detect or adapt to whatever CUDA
+you already have -- so check what your PyTorch install is using first:
 
 .. code-block:: console
 
     python -c "import torch; print(torch.version.cuda)"
 
-If that prints a version outside ``onnxruntime-gpu``'s default -- for example CUDA 11.x, which
-is no longer published to PyPI under the default package name -- use ONNX Runtime's
-`installation selector <https://onnxruntime.ai/getting-started>`_ to get the right command for
-your setup.
-
-You do **not** normally need a separate system CUDA toolkit. ``onnxruntime-gpu`` can reuse the
-CUDA and cuDNN libraries that PyTorch already bundles, provided ``torch`` is imported first --
-which it always is here, since ``lightning_pose.api.Model`` imports ``torch`` at module level.
-On a working PyTorch + CUDA setup, ``pip install onnxruntime-gpu`` is usually all you need.
-
-.. note::
-
-   If the CUDA major versions don't match, ``onnxruntime`` does not raise. It silently drops
-   ``CUDAExecutionProvider`` and runs on CPU instead, which "works" but is drastically slower
-   with no indication why. Lightning Pose checks for this and raises a ``RuntimeError`` when a
-   CUDA device is available but ``CUDAExecutionProvider`` failed to load. If you hit that
-   error, re-check ``torch.version.cuda`` against your installed ``onnxruntime-gpu`` build and
-   reinstall the matching one.
+Then use that CUDA version to select the proper installation option from ONNX Runtime's
+`installation selector <https://onnxruntime.ai/getting-started>`_.
 
 Usage
 ~~~~~
--- a/docs/source/user_guide_advanced/increasing_inference_speed.rst
+++ b/docs/source/user_guide_advanced/increasing_inference_speed.rst
@@ -187,64 +187,131 @@ a no-op.
 ONNX Runtime
 ------------
 
-Export the model's forward pass to ONNX, then run inference through ``onnxruntime``, plugging
-the ONNX session back in as the model's forward pass the same way as above:
+Installation
+~~~~~~~~~~~~
+
+ONNX Runtime is an optional dependency:
+
+.. code-block:: console
+
+    pip install onnxruntime-gpu
+
+``onnxruntime-gpu`` wheels are built against a specific CUDA major version at build time --
+currently CUDA 12.x/13.x for the default PyPI package. The wheel does **not** detect or adapt
+to whatever CUDA you already have, so check what your PyTorch install is using and match it:
+
+.. code-block:: console
+
+    python -c "import torch; print(torch.version.cuda)"
+
+If that prints a version outside ``onnxruntime-gpu``'s default -- for example CUDA 11.x, which
+is no longer published to PyPI under the default package name -- use ONNX Runtime's
+`installation selector <https://onnxruntime.ai/getting-started>`_ to get the right command for
+your setup.
+
+You do **not** normally need a separate system CUDA toolkit. ``onnxruntime-gpu`` can reuse the
+CUDA and cuDNN libraries that PyTorch already bundles, provided ``torch`` is imported first --
+which it always is here, since ``lightning_pose.api.Model`` imports ``torch`` at module level.
+On a working PyTorch + CUDA setup, ``pip install onnxruntime-gpu`` is usually all you need.
+
+.. note::
+
+   If the CUDA major versions don't match, ``onnxruntime`` does not raise. It silently drops
+   ``CUDAExecutionProvider`` and runs on CPU instead, which "works" but is drastically slower
+   with no indication why. Lightning Pose checks for this and raises a ``RuntimeError`` when a
+   CUDA device is available but ``CUDAExecutionProvider`` failed to load. If you hit that
+   error, re-check ``torch.version.cuda`` against your installed ``onnxruntime-gpu`` build and
+   reinstall the matching one.
+
+Usage
+~~~~~
+
+Exporting and loading are separate steps. Export once, then load with ``runtime="onnx"`` for
+every subsequent prediction run:
 
 .. code-block:: python
 
-    import numpy as np
-    import onnxruntime as ort
-    import torch
     from lightning_pose.api import Model
 
+    # export once
     model = Model.from_dir("path/to/model_dir")
-    real_module = model.model
+    model.export("onnx", onnx_precision="fp16")
 
-    resize_h = model.cfg.data.image_resize_dims.height
-    resize_w = model.cfg.data.image_resize_dims.width
-    # shape is (1, num_views, 3, H, W) instead for a multi-view model
-    dummy = torch.randn(1, 3, resize_h, resize_w, device="cuda")
-
-    torch.onnx.export(
-        real_module, dummy, "/path/to/model.onnx",
-        input_names=["images"], output_names=["heatmaps"],
-        dynamic_axes={"images": {0: "batch"}, "heatmaps": {0: "batch"}},
-        opset_version=17, do_constant_folding=True,
+    # load through ONNX Runtime and predict as normal
+    onnx_model = Model.from_dir(
+        "path/to/model_dir", runtime="onnx", onnx_precision="fp16"
     )
+    result = onnx_model.predict_on_video_file("path/to/video.mp4")
 
-    session = ort.InferenceSession(
-        "/path/to/model.onnx", providers=["CUDAExecutionProvider", "CPUExecutionProvider"]
-    )
+The same two steps from the CLI:
 
-    def onnx_forward(images):
-        io_binding = session.io_binding()
-        io_binding.bind_input(
-            name="images", device_type="cuda", device_id=0,
-            element_type=np.float32, shape=tuple(images.shape),
-            buffer_ptr=images.contiguous().data_ptr(),
-        )
-        io_binding.bind_output(name="heatmaps", device_type="cuda", device_id=0)
-        session.run_with_iobinding(io_binding)
-        return torch.from_numpy(io_binding.copy_outputs_to_cpu()[0]).to(images.device)
+.. code-block:: console
 
-    real_module.forward = onnx_forward
-    result = model.predict_on_video_file("path/to/video.mp4")
+    litpose export /path/to/model_dir --runtime onnx --onnx-precision fp16
+    litpose predict /path/to/model_dir /path/to/video.mp4 --runtime onnx --onnx-precision fp16
 
-Requires ``pip install onnxruntime-gpu`` -- make sure the CUDA version matches your PyTorch
-install. Pip's default ``onnxruntime-gpu`` build may target a newer CUDA major version than
-your PyTorch install uses; check Microsoft's ``onnxruntime-cuda-12`` package index if you're on
-CUDA 12.
+Exports are written to a fixed location inside the model directory:
+
+.. code-block:: text
+
+    model_dir/
+    ├── config.yaml
+    ├── tb_logs/.../checkpoints/epoch=214-step=12685-best.ckpt
+    └── exports_onnx/
+        ├── epoch=214-step=12685-best_fp16.onnx
+        └── epoch=214-step=12685-best_fp32.onnx   # only if exported
+
+You never pass export paths yourself -- ``export()`` always writes here and
+``from_dir(runtime="onnx")`` always reads from here. The filename is the checkpoint's own stem
+plus the export precision, so it stays unambiguous which checkpoint an export came from if a
+model directory ever holds exports alongside an updated checkpoint. ``onnx_precision`` may be
+omitted when exactly one export exists for the checkpoint, and is required to disambiguate when
+several do.
+
+.. note::
+
+   ``precision`` and ``onnx_precision`` are different settings.
+
+   * ``precision`` (``"fp32"``/``"fp16"``/``"bf16"``, also ``litpose predict --precision``)
+     controls the autocast precision of an **eager** forward pass. It leaves the weights on
+     disk untouched.
+   * ``onnx_precision`` (``"fp32"``/``"fp16"``, also ``--onnx-precision``) is the weight
+     precision **baked into the exported** ``.onnx`` **file itself**.
+
+   With ``runtime="onnx"``, ``precision`` is ignored -- the exported file's own precision is
+   what runs.
+
+.. note::
+
+   ``runtime="onnx"`` rebinds the model's ``forward`` method to the ONNX session rather than
+   wrapping the module, for the same reason described in the ``torch.compile()`` note above.
+
+.. note::
+
+   ``torch.compile()`` has no effect on an ONNX Runtime session. Combining ``--compile`` with
+   ``--runtime onnx``, or calling ``compile()`` on a model loaded with ``runtime="onnx"``,
+   raises rather than silently doing nothing.
 
 .. _usage_tensorrt:
 
 TensorRT
 --------
 
-TensorRT engines are built through the same ONNX file, using onnxruntime's
-``TensorrtExecutionProvider`` instead of ``CUDAExecutionProvider`` (continuing the snippet
-above):
+TensorRT engines are built from the same ``.onnx`` file that ``model.export("onnx")``
+produces, using onnxruntime's ``TensorrtExecutionProvider`` instead of
+``CUDAExecutionProvider``. There is no ``runtime="tensorrt"`` yet, so the session is built
+by hand:
 
 .. code-block:: python
+
+    import onnxruntime as ort
+
+    from lightning_pose.api import Model
+
+    model = Model.from_dir("path/to/model_dir")
+    onnx_path = model.export("onnx", onnx_precision="fp32")
+    resize_h = model.cfg.data.image_resize_dims.height
+    resize_w = model.cfg.data.image_resize_dims.width
 
     trt_options = {
         "trt_engine_cache_enable": True,
@@ -255,7 +322,7 @@ above):
         "trt_profile_max_shapes": f"images:1x3x{resize_h}x{resize_w}",
     }
     session = ort.InferenceSession(
-        "/path/to/model.onnx",
+        str(onnx_path),
         providers=[
             ("TensorrtExecutionProvider", trt_options),
             "CUDAExecutionProvider",
--- a/lightning_pose/api/model.py
+++ b/lightning_pose/api/model.py
@@ -511,7 +511,7 @@ class Model:
         # Imported lazily: onnxruntime is an optional dependency, and a
         # module-level import would make it mandatory for every eager user.
         try:
-            import onnxruntime as ort
+            import onnxruntime as ort  # type: ignore[import-not-found]
         except ImportError as e:
             raise ImportError(
                 "onnxruntime is required for runtime='onnx' but is not installed. "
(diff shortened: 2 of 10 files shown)


docs/source/user_guide_advanced/increasing_inference_speed.rst
```diff
@@ -187,65 +187,132 @@ a no-op.
...
-Export the model's forward pass to ONNX, then run inference through ``onnxruntime``, plugging
-the ONNX session back in as the model's forward pass the same way as above:
+Installation
+~~~~~~~~~~~~
+
+ONNX Runtime is an optional dependency:
+
+.. code-block:: console
+
+    pip install onnxruntime-gpu
```
themattinthehatt: don't lead with this - a user will likely run `pip install onnxruntime-gpu` before reading further, potentially installing the wrong version. Better to:

1. lead with the following paragraph ("`onnxruntime-gpu` wheels are built...")
2. Then just lead them to the onnx installation selector site and tell them to follow directions there (no need to say "If that prints a version outside..." - just say, "use the CUDA version to select the proper installation option."

docs/source/user_guide_advanced/increasing_inference_speed.rst
```diff
@@ -187,65 +187,132 @@ a no-op.
...
+.. code-block:: console
+
+    python -c "import torch; print(torch.version.cuda)"
+
+If that prints a version outside ``onnxruntime-gpu``'s default -- for example CUDA 11.x, which
+is no longer published to PyPI under the default package name -- use ONNX Runtime's
+`installation selector <https://onnxruntime.ai/getting-started>`_ to get the right command for
+your setup.
+
+You do **not** normally need a separate system CUDA toolkit. ``onnxruntime-gpu`` can reuse the
```
themattinthehatt: can remove this paragraph, since there is nothing actionable here for a user

docs/source/user_guide_advanced/increasing_inference_speed.rst
```diff
@@ -187,65 +187,132 @@ a no-op.
...
+CUDA and cuDNN libraries that PyTorch already bundles, provided ``torch`` is imported first --
+which it always is here, since ``lightning_pose.api.Model`` imports ``torch`` at module level.
+On a working PyTorch + CUDA setup, ``pip install onnxruntime-gpu`` is usually all you need.
+
+.. note::
```
themattinthehatt: remove this note; let's put all of this information in the RuntimeError so it only lives in one place

lightning_pose/api/model.py
```diff
@@ -257,12 +274,22 @@ class Model:
     """Whether ``compile()`` has been called. Guards against double-wrapping
     ``forward`` on repeat calls."""
 
+    _runtime: str = "eager"
```
themattinthehatt: let's make a `_Runtime` `Literal` for better type checking - for now it will have "eager" and "onnx", with "tensorrt" to come later

lightning_pose/api/model.py
```diff
@@ -311,7 +355,17 @@ def from_dir2(
...
+        model = Model(model_dir, config, precision=precision)
+
+        if runtime == "eager":
+            return model
+        elif runtime == "onnx":
+            model._attach_onnx_runtime(onnx_precision)
+            return model
+        else:
+            raise ValueError(
+                f"Unsupported runtime: '{runtime}'. Use 'eager' or 'onnx'."
```
themattinthehatt: can enumerate `_Runtime` rather than hard-coding the runtime options

lightning_pose/api/model.py
```diff
@@ -400,6 +460,239 @@ def _load(self) -> None:
...
+        onnx_path = candidates[0]
+
+        # Imported lazily: onnxruntime is an optional dependency, and a
+        # module-level import would make it mandatory for every eager user.
+        try:
+            import onnxruntime as ort  # type: ignore[import-not-found]
+        except ImportError as e:
+            raise ImportError(
+                "onnxruntime is required for runtime='onnx' but is not installed. "
+                "Install it with `pip install onnxruntime-gpu`, matching the CUDA "
```
themattinthehatt: remove `pip install onnxruntime-gpu` suggestion and just point user directly to the installation section. also, replace this relative path with the actual docs url to make it faster to navigate there.

lightning_pose/api/model.py
```diff
@@ -400,6 +460,239 @@ def _load(self) -> None:
...
+        # on CPU instead, which "works" but is drastically slower with no
+        # indication why. Fail loudly here instead.
+        providers = session.get_providers()
+        if torch.cuda.is_available() and "CUDAExecutionProvider" not in providers:
+            raise RuntimeError(
+                "ONNX Runtime failed to load CUDAExecutionProvider even though a CUDA "
+                "device is available -- inference would silently run on CPU instead. "
+                "This is usually a CUDA/cuDNN major version mismatch between "
+                "onnxruntime-gpu and your installed PyTorch. See the installation "
+                "section of docs/source/user_guide_advanced/increasing_inference_speed.rst "
```
themattinthehatt: same, point directly to a url

themattinthehatt (changes requested):

jjaw123:
All seven addressed.

One thing worth confirming: the errors deep-link to ...increasing_inference_speed.html#onnx-installation. I added an explicit .. _onnx_installation: label and verified Sphinx emits id="onnx-installation" in the built HTML, but that URL hardcodes /en/latest/ as the base — let me know if you'd rather those point somewhere else.

themattinthehatt (approved):

themattinthehatt:
@jjaw123 don't worry about the test failures - I installed ONNX Runtime on the GPU server where we run the unit tests and obviously something is wrong there. debugging needed on my side, not yours. Can you verify these tests are passing on your end?

jjaw123:
Verified on my end — all 41 pass on da870bd, including the 25 GPU-marked export/runtime tests and the 16 that mock onnxruntime, plus 120 CLI tests. Environment: Tesla T4, torch 2.8.0+cu128 (CUDA 12.8), onnxruntime 1.28.0, with CUDAExecutionProvider confirmed active.

themattinthehatt:
ok something is wrong with the PyPI access on the server, I'll need to contact IT. In the meantime I'll make sure these tests skip.

one extra thing - did you need to run a separate `pip install onnx` after `pip install onnxruntime-gpu`? if so, should include that second `pip install onnx` command in the docs (I missed it at first)

jjaw123:
Yes — confirmed empirically. Uninstalled onnx on a throwaway GPU job and re-ran: all 12 export tests fail with torch.onnx.OnnxExporterError: Module onnx is not installed! from torch.onnx.export → _add_onnxscript_fn. onnxruntime-gpu declares only flatbuffers, numpy, packaging, protobuf, so it never comes along. I had it from the earlier TensorRT work, which is why I didn't hit it.

The 22 runtime-only tests passed without it, so onnx is needed to export, not to run an already-exported file — a deployment machine only needs onnxruntime-gpu. Added the second command to the installation section with that distinction.

themattinthehatt (approved):

PR #475, opened 2026-08-08
Add TensorRT export and inference runtime support

### LINKED ISSUE #472: add TensorRT export and inference runtime support
## Background
 
Benchmarking showed TensorRT gives the largest inference speedups of the techniques we've tried — up to 8.32x on the 6-view multiview transformer on an A100 — beating both `torch.compile()` and plain ONNX Runtime in every case tested (see `docs/source/user_guide_advanced/increasing_inference_speed.rst`).
 
This is a follow-up to #470 (ONNX export/runtime, merged in #471). **Please read #471 in full before starting** — it's the actual merged implementation this issue builds on, not just a sketch, and several design questions that would otherwise come up here were already resolved there. The relevant pieces to reuse directly:
 
- `Model._ckpt_stem()` — shared checkpoint-resolution helper. Use this, don't re-derive the checkpoint stem separately.
- `Model.exports_onnx_dir()` — follow this exact pattern for a new `exports_trt_dir()`.
- `Model._runtime` attribute and the `_Runtime` `Literal` type (added during #471's review) — extend the `Literal` to include `"tensorrt"` rather than reintroducing a bare `str`.
- The provider-guard pattern in `_attach_onnx_runtime()` (raises rather than silently falling back) — TensorRT needs an analogous, stricter guard (see below).
- The deferred-import pattern (`import onnxruntime` happens inside `_attach_onnx_runtime()`, not at module level) — do the same for any TensorRT-specific imports, so `tensorrt`/`onnxruntime` stay optional for eager-only and ONNX-only users.
- The dtype-from-session-metadata fix (#471 initially hardcoded `element_type=np.float32` in `onnx_forward`, which broke fp16; it now reads dtype from the session's input metadata). Reuse whatever helper #471 introduced for this rather than hardcoding again.
**Worth considering as a refactor while you're in there:** the `onnx_forward` closure in `_attach_onnx_runtime()` (io_binding setup, dtype detection, CPU/GPU branching) is very likely going to look almost identical for TensorRT — same `InferenceSession` + `io_binding` API, just different `providers`. Consider factoring the shared parts into a helper both `_attach_onnx_runtime()` and the new `_attach_tensorrt_runtime()` can call, rather than duplicating the whole closure. Not required, but flag it in the PR if you decide not to, so it's a deliberate choice.
 
## Directory structure
 
Building on `exports_onnx/` from #471:
 
```
model_dir/
├── config.yaml
├── tb_logs/.../checkpoints/epoch=214-step=12685-best.ckpt
├── exports_onnx/
│   └── epoch=214-step=12685-best_fp16.onnx
└── exports_trt/
    └── epoch=214-step=12685-best_fp16/
        ├── ...engine cache files (onnxruntime-managed)...
        └── trt_metadata.json
```
 
`exports_trt/` needs a subdirectory per checkpoint+precision (not a flat file), because `trt_engine_cache_path` is a directory onnxruntime writes multiple cache files into — cache files for different precisions can't share one directory without risking cache-key collisions.
 
### 1. Add a `"tensorrt"` branch to `Model.export()`
 
TensorRT doesn't re-trace the model from scratch — it builds an engine from the **existing** `.onnx` file produced by `export("onnx", ...)`. This means `export("tensorrt", ...)` should:
 
- Require the matching `.onnx` file to already exist (via `exports_onnx_dir()` + `_ckpt_stem()`), and raise a clear `FileNotFoundError` pointing at `export("onnx", ...)` if it doesn't — don't silently trigger an ONNX export as a side effect.
- Accept `onnx_precision` (which `.onnx` file to build from) and batch-profile bounds.
- Build the engine by running one dummy forward pass through `TensorrtExecutionProvider`, using a **dynamic-shape profile** (min/opt/max) rather than a fixed batch size, so the resulting engine handles a range of batch sizes at inference time — correctness holds across the whole range, but performance is best near `opt`.
```python
def export(
    self,
    runtime: str,
    onnx_precision: str = "fp16",
    max_batch_size: int = 8,
    opt_batch_size: int | None = None,
) -> Path:
    ...
    if runtime == "tensorrt":
        onnx_path = self._find_onnx_export(onnx_precision)  # reuse export-lookup logic from #471
        if onnx_path is None:
            raise FileNotFoundError(
                f"No ONNX export found for onnx_precision='{onnx_precision}'. "
                f"Run model.export('onnx', onnx_precision='{onnx_precision}') first."
            )
 
        opt_batch_size = opt_batch_size or max_batch_size
        stem = self._ckpt_stem()
        cache_dir = self.exports_trt_dir() / f"{stem}_{onnx_precision}"
        cache_dir.mkdir(parents=True, exist_ok=True)
 
        resize_h = self.cfg.data.image_resize_dims.height
        resize_w = self.cfg.data.image_resize_dims.width
        # NOTE: reuse the same singleview/multiview/context dummy-shape branching
        # already implemented in the ONNX export path in #471 -- don't reimplement it.
        shape_suffix = f"3x{resize_h}x{resize_w}"  # adjust per input layout as export()'s ONNX branch does
 
        trt_options = {
            "trt_engine_cache_enable": True,
            "trt_engine_cache_path": str(cache_dir),
            "trt_fp16_enable": onnx_precision == "fp16",
            "trt_profile_min_shapes": f"images:1x{shape_suffix}",
            "trt_profile_opt_shapes": f"images:{opt_batch_size}x{shape_suffix}",
            "trt_profile_max_shapes": f"images:{max_batch_size}x{shape_suffix}",
        }
        session = ort.InferenceSession(
            str(onnx_path),
            providers=[("TensorrtExecutionProvider", trt_options), "CUDAExecutionProvider"],
        )
        if "TensorrtExecutionProvider" not in session.get_providers():
            raise RuntimeError(
                "TensorRT engine build failed to use TensorrtExecutionProvider. "
                "This usually means a missing/mismatched TensorRT install. See the "
                "installation section of docs/source/user_guide_advanced/increasing_inference_speed.rst."
            )
 
        self._write_trt_metadata(cache_dir, max_batch_size, opt_batch_size)
        return cache_dir
```
 
**Open question for you to resolve during implementation, not to guess at up front:** does building the engine require `self._load()` (real torch weights) at all, or can it work directly from the `.onnx` file + `self.cfg` for shape info, skipping weight loading entirely? Given the `.onnx` file already has the trained weights baked in, my guess is no torch-side loading is needed for the `"tensorrt"` branch specifically — but verify this rather than assuming, the same way #471 discovered empirically (not by design) that `torch.onnx.export` needs the separate `onnx` pip package.
 
### 2. Write a `trt_metadata.json` sidecar
 
Unlike the `.onnx` file, a TensorRT engine cache is tied to the exact GPU architecture, TensorRT version, and batch profile it was built with — it is **not portable** across machines. Record this at build time so a later mismatch produces a clear error instead of a confusing onnxruntime failure or an unexpected silent rebuild:
 
```json
{
  "gpu_name": "Tesla T4",
  "tensorrt_version": "10.16.1.11",
  "onnxruntime_version": "1.28.0",
  "onnx_precision": "fp16",
  "max_batch_size": 8,
  "opt_batch_size": 8,
  "built_at": "2026-08-06T12:00:00Z"
}
```
 
`_attach_tensorrt_runtime()` (below) should read this back and, at minimum, warn (not necessarily raise) if the current GPU name doesn't match what's recorded — a mismatch there is exactly the situation from the original benchmarking doc where onnxruntime silently rebuilds the cache on first inference, incurring an unexpected multi-minute stall in what looked like a fast path.
 
### 3. Extend `Model.from_dir` with `runtime="tensorrt"`
 
```python
def _attach_tensorrt_runtime(self, onnx_precision: str | None) -> None:
    stem = self._ckpt_stem()
    cache_dir = self._find_trt_cache(stem, onnx_precision)  # analogous lookup/disambiguation to _find_onnx_export
    if cache_dir is None:
        raise FileNotFoundError(
            f"No TensorRT export found for checkpoint '{stem}'"
            + (f" with onnx_precision='{onnx_precision}'" if onnx_precision else "")
            + f". Run model.export('tensorrt', ...) first."
        )
 
    metadata = json.loads((cache_dir / "trt_metadata.json").read_text())
    if torch.cuda.is_available():
        current_gpu = torch.cuda.get_device_name(0)
        if current_gpu != metadata["gpu_name"]:
            warnings.warn(
                f"TensorRT engine at {cache_dir} was built on '{metadata['gpu_name']}' "
                f"but is running on '{current_gpu}'. onnxruntime will rebuild the engine "
                f"on first inference, which can take several minutes."
            )
    else:
        raise RuntimeError(
            "runtime='tensorrt' requires a CUDA-capable GPU; none is available. "
            "TensorRT has no CPU execution path -- use runtime='onnx' or runtime='eager' instead."
        )
 
    onnx_path = self._find_onnx_export(onnx_precision or metadata["onnx_precision"])
    trt_options = {
        "trt_engine_cache_enable": True,
        "trt_engine_cache_path": str(cache_dir),
        "trt_fp16_enable": metadata["onnx_precision"] == "fp16",
    }
    session = ort.InferenceSession(
        str(onnx_path),
        providers=[("TensorrtExecutionProvider", trt_options), "CUDAExecutionProvider"],
    )
    # Always confirm TensorRT actually engaged -- onnxruntime silently falls back to
    # CUDAExecutionProvider if TensorRT can't load, e.g. a missing library. Unlike
    # ONNX's CUDA-vs-CPU guard, there is no acceptable fallback tier here: if the
    # user asked for runtime="tensorrt", CUDAExecutionProvider alone is a failure,
    # not a degraded-but-working state, so raise rather than warn.
    if "TensorrtExecutionProvider" not in session.get_providers():
        raise RuntimeError(
            "ONNX Runtime failed to load TensorrtExecutionProvider -- inference would "
            "silently run on CUDAExecutionProvider instead, without the TensorRT speedup "
            "you asked for. This is usually a missing/mismatched TensorRT install or a "
            "missing LD_LIBRARY_PATH entry. See the installation section of "
            "docs/source/user_guide_advanced/increasing_inference_speed.rst#tensorrt-installation."
        )
 
    self._load()
    if self.model is None:
        raise RuntimeError('model failed to load; self.model is None after _load()')
    self.model.forward = _build_onnx_forward(session)  # reuse #471's forward-binding helper
    self._runtime = "tensorrt"
```
 
Note the guard here is intentionally **stricter** than ONNX's: ONNX Runtime has a legitimate CPU fallback tier (slower, but correct and expected on a CPU-only machine), so #471's guard only fires when CUDA is available but wasn't used. TensorRT has no CPU path at all, so `runtime="tensorrt"` on a CPU-only machine should fail immediately and clearly, and even on a GPU machine, silently landing on `CUDAExecutionProvider` instead of `TensorrtExecutionProvider` defeats the entire point of asking for this runtime — so that should raise too, not warn.
 
### 4. CLI changes
 
```bash
litpose export /path/to/model_dir --runtime tensorrt --onnx-precision fp16 --max-batch-size 8
litpose predict /path/to/model_dir /path/to/video.mp4 --runtime tensorrt
```
 
```python
export_parser.add_argument("--runtime", choices=["onnx", "tensorrt"], default="onnx")
export_parser.add_argument("--max-batch-size", type=int, default=8)
export_parser.add_argument("--opt-batch-size", type=int, default=None)
 
predict_parser.add_argument("--runtime", choices=["eager", "onnx", "tensorrt"], default="eager")
```
 
Follow #471's pattern of rejecting invalid flag combinations (e.g. `--compile` with `--runtime tensorrt`) inside the CLI before the model is constructed, so the error message matches `compile()`'s own guard rather than surfacing as a raw traceback.
 
## Testing
 
- Unit test: `export("tensorrt")` produces a cache directory and `trt_metadata.json` with the expected fields.
- Unit test: `export("tensorrt")` raises a clear `FileNotFoundError` if the matching `.onnx` export doesn't exist yet.
- Unit test: `from_dir(runtime="tensorrt")` raises `RuntimeError` (not falling back) when `TensorrtExecutionProvider` isn't in `session.get_providers()` — mocked, following #471's pattern for the analogous ONNX provider-guard test.
- Unit test: `from_dir(runtime="tensorrt")` raises immediately when no CUDA device is available, without attempting to build a session.
- Unit test: GPU-name mismatch between `trt_metadata.json` and the current device triggers a warning, not a hard failure.
- Accuracy test: compare predictions from the TensorRT runtime against the eager model on real frames, same style as #471's `test_onnx_matches_eager_predictions`.
- CLI tests: mock-based, following #471's pattern (assert flag parsing and pass-through, not full end-to-end).
- Test on both singleview and multiview models, reusing whatever shape-branching test fixtures #471 already introduced.

## Documentation
 
- Add an installation section for TensorRT to `docs/source/user_guide_advanced/increasing_inference_speed.rst`, following the pattern #471 established for the ONNX install section (including its `.. _onnx_installation:` label convention — add a matching `.. _tensorrt_installation:` label so the `RuntimeError` above can link to it). Cover:
  - The `tensorrt` package needs a version matched to the installed `onnxruntime-gpu` build (e.g. `pip install "tensorrt-cu12<11"` to get TensorRT 10.16.1.11, matching onnxruntime 1.28.0's expected `libnvinfer.so.10` — verify current version numbers empirically rather than trusting this issue's numbers, the same way #471 discovered the separate `onnx` package requirement empirically rather than assuming).
  - `LD_LIBRARY_PATH` needs to include TensorRT's shared libraries at runtime.
  - Whether exporting a TensorRT engine requires anything beyond `onnxruntime-gpu` + `tensorrt` (e.g. does it need the `onnx` package the way ONNX export did? Test this directly rather than assuming either way, given #471's experience.).
- Add `exports_trt/` to the model directory tree in `docs/source/directory_structure_reference/model_dir_structure.rst`, alongside the `exports_onnx/` entry added in #471.
- Document that TensorRT engines are GPU-specific and not portable — a `model_dir` exported on one GPU and copied to a different GPU type will trigger a rebuild (or, per the guard above, a warning) rather than reusing the cache.
- Make the usage docs (for both CLI and API) complete - assume the user *hasn't* read the ONNX docs. So you'll need to walk them through exporting to ONNX first, then running an additional export to TensorRT.

Don't touch dependency pinning beyond what's documented above — exact version numbers should be verified empirically against whatever `onnxruntime-gpu` version is current at implementation time, not hardcoded from this issue. Also don't revisit the `torch.onnx.export` → `dynamo=True` migration flagged as a follow-up in #471; that's unrelated to TensorRT and will be its own issue.

jjaw123:
Closes #472.

Adds TensorRT export and inference runtime support, mirroring the structure from #470/#471 (ONNX export/runtime) and #467/#469 (torch.compile).

## API / CLI

- `Model.export("tensorrt", onnx_precision="fp16", max_batch_size=8, opt_batch_size=None)` builds a TensorRT engine cache from an existing ONNX export (via `onnxruntime`'s `TensorrtExecutionProvider`), written to `exports_trt/<checkpoint_stem>_<onnx_precision>/`.
- `Model.from_dir(model_dir, runtime="tensorrt", onnx_precision="fp16")` loads that engine and rebinds `forward`, same pattern as `runtime="onnx"`.
- `litpose export --runtime tensorrt --onnx-precision fp16 --max-batch-size 8 --opt-batch-size ...` / `litpose predict ... --runtime tensorrt --onnx-precision fp16` on the CLI side.
- A `trt_metadata.json` sidecar records GPU name, TensorRT/onnxruntime versions, and batch profile; `from_dir(runtime="tensorrt")` warns on a GPU-name mismatch, since engine caches are tied to the exact GPU they were built on (unlike the portable `.onnx` file).
- Shared the ONNX/TensorRT `forward`-rebinding closure into one `_build_onnx_forward()` helper used by both runtimes, per the optional suggestion in #470.

## Deviations from the issue's sketch (disclosed, not blockers)

- `_export_tensorrt()` doesn't call `self._load()` and never touches the trained torch weights -- confirmed empirically that the `TensorrtExecutionProvider` builds purely from the `.onnx` file (which already has weights baked in). Input shape for the dynamic-shape batch profile is read directly off the ONNX graph's own input metadata instead of re-deriving it from `self.cfg`/`self.model.do_context`, avoiding a second singleview/multiview/context branching implementation.
- The `onnx` package is required to *create* the `.onnx` file but not to build a TensorRT engine from an already-exported one -- confirmed by uninstalling it on a throwaway job and running the suite. Documented with that split rather than blanket "install onnx and onnxruntime-gpu."
- No CPU fallback tier for `runtime="tensorrt"` at all (raises immediately if no CUDA), stricter than ONNX Runtime's CPU tier -- there's no reasonable degraded mode for a GPU-built engine.

## Accuracy testing -- found something worth flagging

Following #471's precedent (`toy_data_dir`, threshold `<0.1px`) for the accuracy test didn't hold up under investigation, so I dug in rather than just loosening the number:

- **`test_tensorrt_fp32_matches_eager_predictions`** (new): TensorRT at FP32 vs eager is essentially exact -- max deviation 0.0038px across all keypoints, unfiltered. This isolates the export/engine-build pipeline's own correctness from precision noise. Threshold `<0.01px`.
- **`test_tensorrt_fp16_matches_eager_predictions`**: at FP16, an initial run failed at 2.74px max. Root-caused (not just loosened) via a breakdown script: every large deviation traced to keypoints at near-zero eager likelihood (~0.0003, i.e. the heatmap has no real peak) -- the same "peak-flipping" phenomenon already documented for the merged ONNX FP16 path on the full (non-toy) dataset, just showing up here because this toy fixture checkpoint leaves several bodyparts permanently at chance confidence. Filtering to confidently-tracked keypoints (eager likelihood >= 0.1) and repeating the fp16 export+build 3 more times showed the **max** is unstable run-to-run (0.52px / 1.95px / 0.92px) -- TensorRT's kernel autotuning makes engine builds non-deterministic -- while the **mean** stayed stable (~0.2-0.3px) across the same runs. So this test asserts on mean deviation (`<1.0px`, headroom over the observed range) rather than max, with the reasoning in the docstring.

Both findings, plus the "why mean not max" and "why FP16 differs from the FP32 accuracy check earlier on the docs page" reasoning, are also written into `increasing_inference_speed.rst`'s TensorRT section rather than left implicit.

## Testing

- `pyright`: 0 errors (project-wide)
- `ruff check`: clean
- `pre-commit run`: clean
- GPU (T4, `tests/api/test_model.py -q`, no marker filter, to catch collateral damage): 85/85 passed
- CPU (`tests/cli/test_export.py tests/cli/test_predict.py tests/api/test_model.py -m "not gpu"`): 81/81 passed
- Docs: `docutils` clean, full `sphinx-build` clean (verified the `#tensorrt-installation` anchor and both `exports_trt/` mentions render in the built HTML)

## Docs

- New "TensorRT" Installation/Usage subsections in `increasing_inference_speed.rst` (version-matching guidance, `LD_LIBRARY_PATH`, the `onnx`-not-required-to-build finding, the new API usage, the accuracy caveat above).
- `exports_trt/` entry added to `model_dir_structure.rst`.

### DIFF
--- a/docs/source/user_guide_advanced/increasing_inference_speed.rst
+++ b/docs/source/user_guide_advanced/increasing_inference_speed.rst
@@ -287,64 +287,114 @@ several do.
 TensorRT
 --------
 
-TensorRT engines are built from the same ``.onnx`` file that ``model.export("onnx")``
-produces, using onnxruntime's ``TensorrtExecutionProvider`` instead of
-``CUDAExecutionProvider``. There is no ``runtime="tensorrt"`` yet, so the session is built
-by hand:
+.. _tensorrt_installation:
+
+Installation
+~~~~~~~~~~~~~
+
+TensorRT and ``onnxruntime-gpu`` need matching major versions, and ``pip install
+tensorrt`` / ``pip install onnxruntime-gpu`` each default independently to whatever is
+newest -- so they don't necessarily agree with each other, or with your PyTorch install's
+CUDA version. Check what onnxruntime expects before installing TensorRT:
+
+.. code-block:: console
+
+    python -c "import onnxruntime; print(onnxruntime.__version__)"
+
+The combination we verified working is ``onnxruntime-gpu`` 1.28.0 with
+``pip install "tensorrt-cu12<11"`` (lands on TensorRT 10.16.1.11, providing
+``libnvinfer.so.10``). A newer, unpinned ``pip install tensorrt`` can land on TensorRT
+11.x (``libnvinfer.so.11``) instead, which fails at session-creation time with a
+missing-symbol error rather than a clear version-mismatch message.
+
+TensorRT's shared libraries also need to be on ``LD_LIBRARY_PATH`` at runtime:
+
+.. code-block:: bash
+
+    export LD_LIBRARY_PATH=/path/to/site-packages/tensorrt_libs:$LD_LIBRARY_PATH
+
+Building a TensorRT engine starts from the ``.onnx`` file produced by ``model.export(
+"onnx", ...)`` -- the ``onnx`` package is needed to create that file, but not to build
+the engine from it afterward. A machine that only builds/runs TensorRT engines from an
+already-exported ``.onnx`` file needs ``onnxruntime-gpu`` and ``tensorrt`` alone.
+
+Usage
+~~~~~
+
+TensorRT builds its engine from an existing ONNX export, so export to ONNX first if you
+haven't already:
 
 .. code-block:: python
-
-    import onnxruntime as ort
 
     from lightning_pose.api import Model
 
     model = Model.from_dir("path/to/model_dir")
-    onnx_path = model.export("onnx", onnx_precision="fp32")
-    resize_h = model.cfg.data.image_resize_dims.height
-    resize_w = model.cfg.data.image_resize_dims.width
+    model.export("onnx", onnx_precision="fp16")
+    model.export("tensorrt", onnx_precision="fp16", max_batch_size=8)
 
-    trt_options = {
-        "trt_engine_cache_enable": True,
-        "trt_engine_cache_path": "/path/to/trt_cache",
-        "trt_fp16_enable": True,  # set False to retain FP32 precision
-        "trt_profile_min_shapes": f"images:1x3x{resize_h}x{resize_w}",
-        "trt_profile_opt_shapes": f"images:1x3x{resize_h}x{resize_w}",
-        "trt_profile_max_shapes": f"images:1x3x{resize_h}x{resize_w}",
-    }
-    session = ort.InferenceSession(
-        str(onnx_path),
-        providers=[
-            ("TensorrtExecutionProvider", trt_options),
-            "CUDAExecutionProvider",
-            "CPUExecutionProvider",
-        ],
+    trt_model = Model.from_dir(
+        "path/to/model_dir", runtime="tensorrt", onnx_precision="fp16"
     )
-    # Always confirm TensorRT actually engaged -- onnxruntime silently falls back to
-    # CUDAExecutionProvider if TensorRT can't be used, e.g. a missing library.
-    assert "TensorrtExecutionProvider" in session.get_providers()
+    result = trt_model.predict_on_video_file("path/to/video.mp4")
 
-The first inference call builds and autotunes an engine for the shapes in the profile above,
-which can take anywhere from several seconds to a few minutes depending on the model. The
-engine is cached to ``trt_engine_cache_path`` and reused on subsequent runs with the same
-shapes.
+The same from the CLI:
 
-Getting TensorRT installed and working took some real trial and error, mostly around matching
-library versions:
+.. code-block:: console
 
-- ``pip install tensorrt`` and ``pip install onnxruntime-gpu`` can each default to newer
-  CUDA/TensorRT major versions that don't match each other or your PyTorch install. We needed
-  ``pip install "tensorrt-cu12<11"`` (landing on TensorRT 10.16.1.11) to match onnxruntime
-  1.28.0's expected ``libnvinfer.so.10``.
-- TensorRT's shared libraries need to be on ``LD_LIBRARY_PATH`` at runtime, e.g.:
+    litpose export /path/to/model_dir --runtime onnx --onnx-precision fp16
+    litpose export /path/to/model_dir --runtime tensorrt --onnx-precision fp16 --max-batch-size 8
+    litpose predict /path/to/model_dir /path/to/video.mp4 --runtime tensorrt --onnx-precision fp16
 
-  .. code-block:: bash
+``max_batch_size`` sets the upper end of the dynamic-shape batch profile the engine is
+built for -- inference at a larger batch size fails. ``opt_batch_size`` (defaults to
+``max_batch_size``) is the batch size the engine is tuned to run fastest at; correctness
+holds for any batch size in ``[1, max_batch_size]``, but performance is best near
+``opt_batch_size``.
 
-      export LD_LIBRARY_PATH=/path/to/site-packages/tensorrt_libs:$LD_LIBRARY_PATH
+The engine cache is written alongside the ONNX exports:
 
-- Always check ``session.get_providers()`` after creating the session. If TensorRT can't
-  load (e.g. a missing library), onnxruntime silently falls back to
-  ``CUDAExecutionProvider`` rather than raising an error -- your code will still run and
-  produce plausible-looking numbers, just not the ones you think.
+.. code-block:: text
+
+    model_dir/
+    ├── config.yaml
+    ├── tb_logs/.../checkpoints/epoch=214-step=12685-best.ckpt
+    ├── exports_onnx/
+    │   └── epoch=214-step=12685-best_fp16.onnx
+    └── exports_trt/
+        └── epoch=214-step=12685-best_fp16/
+            ├── trt_metadata.json
+            └── ...                          # engine cache files (managed by onnxruntime)
+
+The first call to ``model.export("tensorrt", ...)`` builds and autotunes the engine, which
+can take anywhere from several seconds to a few minutes. ``trt_metadata.json`` records the
+GPU name, TensorRT/onnxruntime versions, and batch profile the engine was built with --
+``Model.from_dir(..., runtime="tensorrt")`` reads this back and warns if the current GPU
+doesn't match, since (unlike an ``.onnx`` file) a built engine is tied to the exact GPU
+architecture it was built on and is not portable across machines.
+
+.. note::
+
+   Building always requires a real GPU with a working TensorRT/onnxruntime install --
+   there is no CPU fallback tier for ``runtime="tensorrt"``. If
+   ``TensorrtExecutionProvider`` can't load, this raises rather than silently running on
+   ``CUDAExecutionProvider`` or CPU (unlike plain ONNX Runtime, which does have a CPU
+   tier -- see the note in the ONNX Runtime section above about checking
+   ``session.get_providers()``).
+
+.. note::
+
+   The accuracy check earlier on this page used TensorRT at FP32. At FP16, quantization
+   can shift the heatmap's predicted peak by a few pixels on keypoints the model has no
+   real opinion about -- a near-uniform heatmap, likelihood near the noise floor -- a
+   "peak-flipping" effect that happens under any precision or kernel change (it
+   reproduces on the already-merged ONNX FP16 path too, given a real dataset rather than
+   a handful of hand-picked frames) and isn't specific to TensorRT. On
+   confidently-tracked keypoints it stays small: repeated FP16 engine builds during
+   testing measured mean deviation around 0.2-0.3px, though TensorRT's own build-time
+   autotuning adds some run-to-run variance on top -- engine builds aren't perfectly
+   deterministic, so don't expect bit-identical predictions from two separately-built
+   engines either.
+
 
 .. _caveats:
 
--- a/docs/source/user_guide_advanced/increasing_inference_speed.rst
+++ b/docs/source/user_guide_advanced/increasing_inference_speed.rst
@@ -307,10 +307,14 @@ The combination we verified working is ``onnxruntime-gpu`` 1.28.0 with
 11.x (``libnvinfer.so.11``) instead, which fails at session-creation time with a
 missing-symbol error rather than a clear version-mismatch message.
 
+For the full installation matrix (which TensorRT version pairs with which CUDA/cuDNN), see NVIDIA's `TensorRT install guide <https://docs.nvidia.com/deeplearning/tensorrt/latest/installing-tensorrt/installing.html>`_. The ``<11`` pin above matches what onnxruntime-gpu's TensorRT execution provider currently requires. If your onnxruntime-gpu version differs from the one this doc was written against, check onnxruntime's own `TensorRT EP requirements table <https://onnxruntime.ai/docs/execution-providers/TensorRT-ExecutionProvider.html#requirements>`_ first -- it's the maintained source of truth for onnxruntime-gpu <-> TensorRT <-> CUDA compatibility, including versions released after this page was written.
+
 TensorRT's shared libraries also need to be on ``LD_LIBRARY_PATH`` at runtime:
 
 .. code-block:: bash
 
+    # find your site-packages path:
+    #   python -c "import tensorrt_libs, os; print(os.path.dirname(tensorrt_libs.__file__))"
     export LD_LIBRARY_PATH=/path/to/site-packages/tensorrt_libs:$LD_LIBRARY_PATH
 
 Building a TensorRT engine starts from the ``.onnx`` file produced by ``model.export(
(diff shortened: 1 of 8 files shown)


docs/source/user_guide_advanced/increasing_inference_speed.rst
```diff
@@ -287,64 +287,114 @@ several do.
...
+
+.. code-block:: console
+
+    python -c "import onnxruntime; print(onnxruntime.__version__)"
+
+The combination we verified working is ``onnxruntime-gpu`` 1.28.0 with
+``pip install "tensorrt-cu12<11"`` (lands on TensorRT 10.16.1.11, providing
+``libnvinfer.so.10``). A newer, unpinned ``pip install tensorrt`` can land on TensorRT
+11.x (``libnvinfer.so.11``) instead, which fails at session-creation time with a
+missing-symbol error rather than a clear version-mismatch message.
```
themattinthehatt, line 308: is there a tensorrt installation page we can point to that will have more complete instructions? what should someone do if they have `onnxruntime-gpu=1.20`? what about in the future if they have `onnxruntime-gpu=1.31`?

docs/source/user_guide_advanced/increasing_inference_speed.rst
```diff
@@ -287,64 +287,114 @@ several do.
...
+
+TensorRT's shared libraries also need to be on ``LD_LIBRARY_PATH`` at runtime:
 
-    import onnxruntime as ort
+.. code-block:: bash
+
+    export LD_LIBRARY_PATH=/path/to/site-packages/tensorrt_libs:$LD_LIBRARY_PATH
```
themattinthehatt, line 318: leave a comment telling a user how to find their specific `/path/to/site-packages/`

docs/source/user_guide_advanced/increasing_inference_speed.rst
```diff
@@ -287,64 +287,114 @@ several do.
...
-    }
-    session = ort.InferenceSession(
-        str(onnx_path),
-        providers=[
-            ("TensorrtExecutionProvider", trt_options),
-            "CUDAExecutionProvider",
-            "CPUExecutionProvider",
-        ],
+    model.export("onnx", onnx_precision="fp16")
+    model.export("tensorrt", onnx_precision="fp16", max_batch_size=8)
```
themattinthehatt: increase the example `max_batch_size` to something bigger like 32, since I imagine many users will just copy this line and bigger batch size will probably lead to faster inference overall (especially when combined with the model speedups)

docs/source/user_guide_advanced/increasing_inference_speed.rst
```diff
@@ -287,64 +287,114 @@ several do.
...
-    # CUDAExecutionProvider if TensorRT can't be used, e.g. a missing library.
-    assert "TensorrtExecutionProvider" in session.get_providers()
+    result = trt_model.predict_on_video_file("path/to/video.mp4")
+
+The same from the CLI:
+
+.. code-block:: console
+
+    litpose export /path/to/model_dir --runtime onnx --onnx-precision fp16
+    litpose export /path/to/model_dir --runtime tensorrt --onnx-precision fp16 --max-batch-size 8
```
themattinthehatt: increase batch size

lightning_pose/api/model.py
```diff
@@ -76,7 +78,7 @@
 # Inference runtimes accepted by Model.from_dir(runtime=...). "tensorrt" is a
```
themattinthehatt: can remove the second sentence that starts with `"tensorrt" is a planned addition..."

themattinthehatt (changes requested):

themattinthehatt (approved):

PR #477, opened 2026-08-12
Add PyNvVideoCodec decoder backend for video prediction (--decoder flag)

### LINKED ISSUE #476: PyNvVideoCodec inference dataloader as an alternative to DALI
## Motivation

`litpose predict` on video currently always routes through DALI (`PrepareDALI` / `LitDaliWrapper` in `lightning_pose/data/dali.py`). For inference specifically, DALI's randomized/multiview-sync/imgaug-augmentation machinery is unused — `predict` already runs with shuffle and imgaug off. A decoder built on [PyNvVideoCodec](https://developer.nvidia.com/pynvvideocodec) (NVDEC-backed, random-access-by-index) can replicate the much smaller subset actually needed for inference — sequential (optionally multiview-synced) frame reading, context-model overlapping windows, GPU resize+normalize, and bbox-crop — while giving simpler control over context-window construction than DALI's `step`/`pad_sequences` machinery.

This should be a drop-in alternate backend selected at predict time, fully compatible with the existing `--runtime`/`--compile` (eager / torch.compile / ONNX) axis, since it only replaces video ingestion, not the model or `Trainer.predict()` call.

## Scope

`litpose predict` on video inputs only. Image and CSV prediction (`predict_on_label_csv`) use a plain `DataModule` and never touch DALI/video decoding — out of scope.

## CLI

- New flag: **`--decoder`**, choices `dali` / `pynvvc`.
- Default: auto-select `pynvvc` if available (package installed **and** GPU/driver actually support NVDEC — see "Install & default-selection guard" below), else fall back to `dali`. Flag stays optional.
- If the user explicitly passes `--decoder pynvvc` and it isn't usable, fail fast in the CLI handler with an actionable message (mirrors the existing `--compile` + `--runtime onnx` guard in `predict.py:183`), rather than surfacing a raw NVIDIA exception.
- Print which decoder backend is in use. Put this print in `predict_video()` (`lightning_pose/utils/predictions.py`) rather than duplicating it in the CLI and Model API, so every caller (CLI, direct `Model` API/notebook use) gets it for free.

## Model API

- New per-call kwarg on `Model.predict_on_video_file` and `Model.predict_on_video_file_multiview`.
- Threaded through to `predict_video()` in `lightning_pose/utils/predictions.py`, same as `bbox_file` already is.

## Branch point

`predict_video()` (`lightning_pose/utils/predictions.py:412`) is the single place `PrepareDALI` is constructed today. Add a branch there: `dali` → existing `PrepareDALI`/`vid_pred_class()` path unchanged; `pynvvc` → new path. `PredictionHandler` and the `trainer.predict(...)` call do not change.

## New file: `lightning_pose/data/pynvvc.py`

Mirrors `dali.py`'s two-phase construction (`PrepareDALI.__init__` validates/precomputes, `__call__` builds the iterator). Needs to produce `UnlabeledBatchDict` / `MultiviewUnlabeledBatchDict` exactly as consumed by `predict_step` in `heatmap_tracker.py`/`heatmap_tracker_mhcrnn.py`/`heatmap_tracker_multiview.py`/`regression_tracker.py`:

- `frames`: FCHW layout, `[0,1]`-scaled then ImageNet-normalized.
- `transforms`: for prediction (no `dlc`/`dlc-top-down` imgaug), this is just the sentinel `np.array([-1])` per `dali.py:180-182` — no rotation/scale matrix to reconstruct. Augmentation-transform replication only matters for training, which is out of scope.
- `bbox`: full-frame bbox by default, or actual per-frame bbox in crop mode.
- Context models: replicate the 5-frame overlapping-window behavior and the "pad last batch by repeating the final frame" behavior so shapes stay static across steps (see torch.compile note below) and `PredictionHandler` still gets full tensors matched to the true frame count.

## Extract bbox-crop logic into `lightning_pose/data/bboxes.py`

`_apply_bbox_crop` (`dali.py:332-396`) is currently a method coupled to `LitDaliWrapper`'s `self.bbox_df`/`self.do_context`/`self.resize_dims`/`self._frame_idx`. To avoid forking ~65 lines of crop/clamp/resize logic (including the edge-clamping fix for negative x/y from `create_bbox`) into a second class, pull the per-frame crop+resize math into a free function in `bboxes.py` — which is already the module for bbox coordinate-space transforms (`frame_to_model`, `model_to_frame_batch`, etc.):

```python
def crop_and_resize_frames(
    frames: Float[torch.Tensor, "seq_len rgb h w"],
    bbox_rows: pd.DataFrame,
    resize_dims: list[int],
) -> tuple[Float[torch.Tensor, "seq_len rgb h_out w_out"], Float[torch.Tensor, "seq_len xyhw"]]:
```

Takes pre-sliced/padded bbox rows; leaves cursor/step bookkeeping (`_frame_idx`, `seq_len - 4` context step for DALI vs. pynvvc's windowing) in each backend's wrapper. Both `LitDaliWrapper._apply_bbox_crop` and the new pynvvc wrapper call this shared function.

## Install & default-selection guard

- Add `pynvvideocodec` to `pyproject.toml`'s unconditional `dependencies` list (same treatment as `nvidia-dali-cuda110` at line 54), gated by the same `sys_platform == 'linux' and platform_machine == 'x86_64'` marker. Package is ~245 KB compressed — no install-size concern.
- GPU/driver reality: NVDEC required (Turing/Ampere/Ada/Hopper/Blackwell only — Pascal and older unsupported), Linux driver ≥ 530.41.03. `import pynvvideocodec` will succeed regardless of GPU generation or driver age — the failure only surfaces when a decoder is actually constructed. **`is_pynvvc_available()` must attempt a minimal decoder construction inside its try/except, not just the import**, or auto-selection will pick `pynvvc` on an unsupported machine and crash on the first video with a raw NVIDIA error instead of falling back to DALI. One-time cost per predict call, not per-batch — cheap enough to do properly.
- Follow the CPU-only-install lazy-import discipline already documented in CLAUDE.md for `dali.py`: `pynvvideocodec` import stays inside function bodies in `pynvvc.py`, never at module top level, marked with the same `# lazy: avoids ImportError on cpu-only installs` convention.

## CUDA stream synchronization — real risk, needs explicit testing

PyNvVideoCodec hands back decoded frames as DLPack buffers (`torch.from_dlpack(frame)`), decoded asynchronously on the decoder's own CUDA stream. If the model's forward pass (default stream, or a `torch.compile`/cudagraphs side-stream) consumes that tensor without the consumer stream waiting on the decode-complete event, the buffer can be read before NVDEC finishes writing it. This fails silently — no crash, no NaN, just quietly wrong pixels feeding a plausible-looking keypoint prediction.

- Check first whether the decoder can be constructed against an explicit external stream handle (pass `torch.cuda.current_stream()`'s stream in) — if supported, that's strictly better than per-batch `synchronize()` calls: no sync cost, no risk of a missed call on a new code path.
- Test plan: standalone repro script, decode → `torch.from_dlpack` → run under each of the three compute backends (eager, `torch.compile`, ONNX Runtime CUDA EP) separately, since stream handling differs per backend and a sync bug can be backend-specific (cudagraphs capture is especially stream-order-sensitive). Run predictions twice — once with an explicit sync/stream-wait inserted right after the DLPack conversion, once without — and diff the output keypoint CSVs. Also test under concurrent GPU load (two `predict_video` calls at once, or decoding while `--compile` is still warming up) — divergence that only appears under load is the race; a small consistent pixel-error offset between pynvvc and DALI outputs is expected and fine (different resize kernels, not a bug).

## torch.compile / ONNX / TensorRT compatibility

Should be automatic since `predict_step` and the `trainer.predict()` call site are unchanged. One thing the pynvvc wrapper must get right: DALI already pads the last batch to full `seq_len` for context models so shapes stay static across steps; the new "pad by repeating final frame" logic must produce identical shapes, or torch.compile will eat a recompilation stall on the final batch of every video.

## Docs

Make explicit in the inference-speed doc page that `--decoder` (video ingestion: dali/pynvvc) and `--runtime`/`--compile` (model execution: eager/onnx/compiled) are independent axes — show at least one example combining them (`--decoder pynvvc --runtime onnx`), plus a short table of valid combinations (all valid except the existing `--compile` + `--runtime onnx` restriction).

## Files touched

- [ ] `lightning_pose/cli/commands/predict.py` — `--decoder` flag
- [ ] `lightning_pose/api/model.py` — new kwarg on `predict_on_video_file`/`_multiview`
- [ ] `lightning_pose/utils/predictions.py` — branch in `predict_video()`, backend print
- [ ] `lightning_pose/data/pynvvc.py` — new wrapper + `is_pynvvc_available()`
- [ ] `lightning_pose/data/bboxes.py` — extracted `crop_and_resize_frames`
- [ ] `lightning_pose/data/dali.py` — `LitDaliWrapper._apply_bbox_crop` calls the shared helper
- [ ] `pyproject.toml` — unconditional dependency, platform-gated
- [ ] `tests/data/test_pynvvc.py` — mirrors `tests/data/test_dali.py`, gpu-marked
- [ ] inference-speed doc page — `--decoder` flag, independence from `--runtime`/`--compile`

jjaw123:
Closes #476.

Adds a `--decoder` option (`dali` / `pynvvc`) for video prediction, mirroring the
structure of #470/#471 (ONNX export/runtime). Follows up on the decode-speed
benchmark confirming PyNvVideoCodec (direct NVIDIA Video Codec SDK / NVDEC bindings)
decodes **8.76x faster** than DALI's GPU video reader on a T4 (3735.7 vs 426.5 fps,
7 timed runs) -- this PR turns that into a selectable inference-time backend.

## API / CLI

- `Model.predict_on_video_file(decoder="pynvvc")` / `predict_on_video_file_multiview(decoder=...)`
  accept `"dali"`, `"pynvvc"`, or `None`.
- `litpose predict ... --decoder pynvvc` / `--decoder dali` on the CLI.
- `decoder=None` (the default) auto-selects `pynvvc` when it's actually usable on the
  current machine for the given video (checked via `is_pynvvc_available()`, which
  probes with a real decoder construction rather than trusting a bare `import`
  succeeding), falling back to `dali` otherwise. Explicitly requesting
  `decoder="pynvvc"` on a machine where it isn't usable raises a clear `RuntimeError`
  instead of silently falling back.
- `--decoder` is fully independent of `--runtime`/`--compile` -- it only controls
  video ingestion, not model execution.
- Shared the bbox-crop math (`LitDaliWrapper._apply_bbox_crop`'s core logic) out into
  `lightning_pose/data/bboxes.py::crop_and_resize_frames()`, used by both `dali.py`
  and the new `pynvvc.py`, rather than forking ~65 lines into a second class.

## New file: `lightning_pose/data/pynvvc.py`

`PreparePynvvc`/`LitPynvvcWrapper`, mirroring `dali.py`'s `PrepareDALI`/`LitDaliWrapper`
two-phase construction, but predict-only (no shuffle, no imgaug -- `litpose predict`
never uses either). Handles single-view + multiview (lockstep sequential reads), base
and context (5-frame overlapping window) model types, bbox-crop mode, and last-batch
padding (repeats the final frame so shape stays static for torch.compile).

## Design note flagged for review (not resolved)

`PreparePynvvc` reuses `cfg.dali.base/context.predict.sequence_length` rather than
introducing a parallel `cfg.pynvvc` config section, since the windowing semantics are
decoder-agnostic. Open question whether a separate config section is actually wanted.

## CUDA stream synchronization

`PyNvVideoCodec.SimpleDecoder` writes frames asynchronously on its own CUDA stream;
`LitPynvvcWrapper.__next__` guards against reading not-yet-finished decodes with a
blanket `torch.cuda.current_stream().synchronize()` per batch. Safe, but possibly not
optimal -- passing an explicit stream handle into the decoder constructor (if
supported) would avoid the sync cost. Flagged as a follow-up, not done here.

## Testing

- `tests/data/test_pynvvc.py` (28 tests, all GPU-free): `is_pynvvc_available()`
  (including the "PyNvVideoCodec not installed" and "decoder construction fails"
  paths, both previously untested per the code's own docstring), `PreparePynvvc`
  validation/`num_iters` math, and `LitPynvvcWrapper`'s `_read_window`/
  `_resize_normalize`/`__next__` -- built via `object.__new__` + a fake list-based
  decoder (works because `torch.from_dlpack()` accepts anything implementing
  `__dlpack__`, including CPU tensors) + a mocked `torch.cuda.current_stream()`.
  Also closes a previously-untested edge case: `_read_window`'s fully-out-of-bounds
  branch.
- `tests/cli/test_predict.py`: 4 new tests for the `--decoder` argparse flag
  (default/pynvvc/dali/rejects-unknown), plus existing hand-built `Namespace` test
  fixtures updated for the new `decoder` field.
- GPU (T4, `tests/api/test_model.py -q`, no marker filter): 61/61 passed, including
  all three `TestPredictOnVideoFile` tests -- confirms the auto-select/fallback
  wiring works end-to-end on real hardware, not just at the unit level.
- `pyright`: 0 errors. `ruff check`: clean. `pre-commit run`: clean.
- Docs: `sphinx-build` clean, new section's cross-reference link verified resolving
  correctly in the built HTML.

## Docs

New "PyNvVideoCodec (Video Decoding)" section in `increasing_inference_speed.rst`
(install/usage, GPU/driver requirements, the 8.76x decode-throughput number with an
explicit caveat that it's decode-only, not end-to-end), cross-referenced from the
page's existing "isolated forward-pass, not end-to-end" caveat.

## Not done / follow-ups

- `is_pynvvc_available()`'s "returns False safely" path is unit-tested via mocking,
  but not verified against a real unsupported GPU/driver combination.
- The CUDA stream synchronization approach (see above) hasn't been checked against
  an explicit-stream-handle alternative.

### DIFF
--- a/docs/source/user_guide_advanced/increasing_inference_speed.rst
+++ b/docs/source/user_guide_advanced/increasing_inference_speed.rst
@@ -346,6 +346,61 @@ library versions:
   ``CUDAExecutionProvider`` rather than raising an error -- your code will still run and
   produce plausible-looking numbers, just not the ones you think.
 
+.. _pynvvc_decoding:
+
+PyNvVideoCodec (Video Decoding)
+--------------------------------
+
+Everything above speeds up the *model's forward pass*. This section covers a separate,
+complementary axis: how the video frames themselves get decoded off disk before they ever
+reach the model. By default, ``litpose predict`` decodes video with DALI's GPU video reader.
+PyNvVideoCodec (direct NVIDIA Video Codec SDK / NVDEC bindings) is an alternative decoder
+backend that talks to the hardware decoder more directly, skipping DALI's pipeline/graph
+framework overhead.
+
+On a T4, decoding the same video with PyNvVideoCodec measured **8.76x faster** than DALI's GPU
+reader (3735.7 vs 426.5 fps, 7 timed runs). This is a decode-throughput-only number, not an
+end-to-end ``litpose predict`` speedup -- how much it moves the needle overall depends on how
+much of your pipeline's time is spent decoding vs. running the model (see the data-loading
+caveat below).
+
+Installation
+~~~~~~~~~~~~
+
+.. code-block:: bash
+
+    pip install PyNvVideoCodec
+
+Requires an NVIDIA GPU from the Turing generation or newer (Turing/Ampere/Ada/Hopper/Blackwell)
+and driver version 530.41.03 or newer on Linux. If PyNvVideoCodec can't decode a given video on
+the current machine (unsupported GPU generation, driver too old, package not installed, or an
+unsupported video format), ``litpose predict`` automatically falls back to DALI rather than
+erroring -- see below.
+
+Usage
+~~~~~
+
+``--decoder`` is independent of ``--runtime``/``--compile`` above -- it only controls how video
+frames are read, not how the model runs on them. From the CLI:
+
+.. code-block:: console
+
+    litpose predict /path/to/model_dir /path/to/video.mp4 --decoder pynvvc
+    litpose predict /path/to/model_dir /path/to/video.mp4 --decoder dali
+
+Or from the API:
+
+.. code-block:: python
+
+    model = Model.from_dir("path/to/model_dir")
+    result = model.predict_on_video_file("path/to/video.mp4", decoder="pynvvc")
+
+Omitting ``--decoder`` (or passing ``decoder=None``) auto-selects PyNvVideoCodec when it's
+usable on the current machine for the given video, falling back to DALI otherwise -- no error,
+just a quieter/slower run. Explicitly requesting ``--decoder pynvvc`` on a machine where it
+isn't usable raises a clear error instead of silently falling back, so you know your run isn't
+using the backend you asked for.
+
 .. _caveats:
 
 Caveats
@@ -353,7 +408,7 @@ Caveats
 
 - **These are isolated forward-pass numbers, not end-to-end** ``litpose predict`` **timings.**
   Real inference also includes data loading (DALI) and postprocessing. These forward-pass
-  gains aren't guaranteed to translate 1:1 into end-to-end speedups as-is.
+  gains aren't guaranteed to translate 1:1 into end-to-end speedups as-is. See :ref:`PyNvVideoCodec <pynvvc_decoding>` above for a way to speed up the data-loading half of that gap.
 - **cuDNN TF32 was left on for the ResNet50 eager-FP32 baseline** (only matmul TF32 was
   disabled), so it's not a fully strict FP32 number -- doesn't change the direction of any
   result here.
--- a/docs/source/user_guide_advanced/increasing_inference_speed.rst
+++ b/docs/source/user_guide_advanced/increasing_inference_speed.rst
@@ -5,12 +5,14 @@ Increasing Inference Speed
 ##########################
 
 In addition to :ref:`running inference at reduced precision <mixed_precision>`, Lightning Pose
-models can be accelerated further with three additional techniques:
+models can be accelerated further with additional techniques that speed up model processing --
 `torch.compile() <https://pytorch.org/docs/stable/generated/torch.compile.html>`_,
 `ONNX Runtime <https://onnxruntime.ai/>`_, and
-`TensorRT <https://developer.nvidia.com/tensorrt>`_. This page benchmarks all four options
-together (including eager FP16 for reference) and walks through how to use each one, assuming
-you've already trained a model with ``litpose train``.
+`TensorRT <https://developer.nvidia.com/tensorrt>`_ -- and with hardware-accelerated video
+decoding -- :ref:`PyNvVideoCodec <pynvvc_decoding>`. This page benchmarks the model-processing
+techniques together (including eager FP16 for reference) and walks through how to use each one,
+assuming you've already trained a model with ``litpose train``; see the
+:ref:`PyNvVideoCodec section <pynvvc_decoding>` below for the video-decoding benchmarks and usage.
 
 **TL;DR**
 
@@ -24,10 +26,6 @@ you've already trained a model with ``litpose train``.
   max deviation was under 0.08px in every case, consistent with ordinary floating-point kernel
   differences and far below any dataset's typical pixel error (see
   :ref:`Accuracy check <accuracy_check>` below).
-- **These numbers are isolated forward-pass speedups** (no data loading), same methodology as
-  the mixed-precision speed numbers. See the :ref:`caveat <caveats>` at the bottom of this page
-  about the gap between forward-pass and end-to-end ``litpose predict`` speed.
-
 Overview
 ========
 
@@ -43,6 +41,13 @@ Overview
 - **TensorRT** -- same ONNX export, but run through onnxruntime's ``TensorrtExecutionProvider``,
   which builds an autotuned, hardware-specific inference engine. Most setup, biggest gains. See
   :ref:`Usage: TensorRT <usage_tensorrt>`.
+- **PyNvVideoCodec** -- hardware-accelerated video *decoding* via direct NVIDIA Video Codec SDK
+  bindings, as an alternative to DALI's video reader. Independent of the model-processing
+  techniques above -- see :ref:`PyNvVideoCodec (Video Decoding) <pynvvc_decoding>`.
+- **These techniques are complementary and can be combined** -- e.g. torch.compile() or
+  TensorRT for the model together with PyNvVideoCodec for decoding -- for the best end-to-end
+  throughput. Not yet benchmarked together on this page; a combined benchmark is planned for a
+  follow-up PR.
 
 Results
 =======
--- a/docs/source/user_guide_advanced/increasing_inference_speed.rst
+++ b/docs/source/user_guide_advanced/increasing_inference_speed.rst
@@ -364,6 +364,18 @@ end-to-end ``litpose predict`` speedup -- how much it moves the needle overall d
 much of your pipeline's time is spent decoding vs. running the model (see the data-loading
 caveat below).
 
+.. note::
+
+   ``PreparePynvvc`` (the class backing ``--decoder pynvvc``) reads its window size from
+   the existing ``cfg.dali`` config section --
+   ``dali.base.predict.sequence_length`` / ``dali.context.predict.sequence_length`` --
+   rather than a separate ``cfg.pynvvc`` section. The name is a little confusing, since
+   this setting also governs how many frames PyNvVideoCodec reads per batch, but the
+   windowing semantics (frames per iteration, overlap for context/MHCRNN models) are
+   identical between the two decoder backends, so one config value covers both rather
+   than keeping two in sync. If you only ever use ``--decoder pynvvc`` and never DALI,
+   you still configure the window size via ``dali.*.predict.sequence_length``.
+
 Installation
 ~~~~~~~~~~~~
 
--- a/docs/source/user_guide_advanced/increasing_inference_speed.rst
+++ b/docs/source/user_guide_advanced/increasing_inference_speed.rst
@@ -413,6 +413,60 @@ just a quieter/slower run. Explicitly requesting ``--decoder pynvvc`` on a machi
 isn't usable raises a clear error instead of silently falling back, so you know your run isn't
 using the backend you asked for.
 
+Decode speed: L4 and A100, single- and multi-view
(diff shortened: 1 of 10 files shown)


docs/source/user_guide_advanced/increasing_inference_speed.rst
themattinthehatt, line 7: update this paragraph; we want to make it clear that "Lightning Pose models can be accelerated further with additional techniques that speed up model processing - `torch.compile() <https:...`... - and hardware-acclerated video decoding - `PyNvVideoCodec <https:...`."

themattinthehatt: can remove this now, don't want to clutter the tl;dr

themattinthehatt: add point about PyNvVideoCodec speedups

add point that model speedup techniques can be combined with hardware-acclerated video decoding for best performance (we don't explicitly show this yet, we'll come back to this in a later PR I think)

themattinthehatt, line 51: add PyNvVideoCodec point/links

lightning_pose/api/model.py
```diff
@@ -75,8 +75,14 @@
 # autocast precision used for eager inference.
 _OnnxPrecision = Literal["fp32", "fp16"]
 
-# Inference runtimes accepted by Model.from_dir(runtime=...).
-_Runtime = Literal["eager", "onnx", "tensorrt"]
+# Inference runtimes accepted by Model.from_dir(runtime=...). "tensorrt" is a
+# planned addition that will consume the same exported .onnx file through a
+# different execution provider.
+_Runtime = Literal["eager", "onnx"]
```
themattinthehatt: oops my merge from main messed this up, my apologies! please add "tensorrt" back to the list and delete the second sentence again

lightning_pose/utils/predictions.py
```diff
@@ -396,6 +396,7 @@ def predict_video(
     model: Model,
     output_pred_file: str | None = None,
     progress_file: Path | None = None,
+    decoder: Literal["dali", "pynvvc"] | None = None,
```
themattinthehatt: move this literal to `_Decoder` variable at top of file

lightning_pose/utils/predictions.py
```diff
@@ -406,6 +407,7 @@ def predict_video(
     model: Model,
     output_pred_file: list[str] | None = None,
     progress_file: Path | None = None,
+    decoder: Literal["dali", "pynvvc"] | None = None,
```
themattinthehatt: use `_Decoder` variable

lightning_pose/utils/predictions.py
```diff
@@ -414,6 +416,7 @@ def predict_video(
     model: Model,
     output_pred_file: str | list[str] | None = None,
     progress_file: Path | None = None,
+    decoder: Literal["dali", "pynvvc"] | None = None,
```
themattinthehatt: use `_Decoder` variable

lightning_pose/utils/predictions.py
```diff
@@ -473,20 +480,56 @@ def predict_video(
...
+        )
+    else:  # pynvvc
+        from lightning_pose.data.pynvvc import (
+            PreparePynvvc,  # avoids ImportError on cpu-only installs
+        )
+        vid_pred_class = PreparePynvvc(
+            model_type=model_type,
+            filenames=filenames,
+            resize_dims=resize_dims,
+            dali_config=model.config.cfg.dali,
```
themattinthehatt: move below `model_type` arg to parallel structure of the `PrepareDALI` call

tests/data/test_pynvvc.py
```diff
@@ -0,0 +1,396 @@
...
+        with patch.dict(sys.modules, {'PyNvVideoCodec': mock_nvc}):
+            assert is_pynvvc_available('/fake/video.mp4') is False
+
+    def test_returns_false_when_package_not_installed(self):
+        """sys.modules[name] = None forces the same ImportError a real missing install would."""
+        with patch.dict(sys.modules, {'PyNvVideoCodec': None}):
+            assert is_pynvvc_available('/fake/video.mp4') is False
+
+
+class TestPreparePynvvc:
```
themattinthehatt, line 44: add version of `test_multiview_synchronized_frames` from `TestPrepareDALI`, this one is important

lightning_pose/data/bboxes.py
```diff
@@ -284,3 +286,58 @@ def model_to_frame_batch(
         model_keypoints_ = norm_to_frame(model_keypoints_, batch_dict['bbox'])
     # return new keypoints, reshaped to (batch, num_targets)
     return model_keypoints_.reshape((-1, num_targets))
+
+
+def crop_and_resize_frames(
```
themattinthehatt, line 291: add tests for this function

PR #480, opened 2026-08-14
Fix multiview ONNX/TensorRT export; unify inference-speed docs into one end-to-end benchmark figure

jjaw123:
## Summary

Two related pieces:

**1. Multiview ONNX/TensorRT export fix.** The 6-view multiview transformer model failed to export to ONNX (`AttributeError: 'Tensor' object has no attribute 'keys'`) because `Model.export()` passes a raw tensor to `torch.onnx.export()` for tracing, but `HeatmapTrackerMultiviewTransformer.forward()` unconditionally expected a dict batch. A second, related bug surfaced once export succeeded: `predict_step()` passed the raw batch dict straight to `forward()` without unwrapping, which broke prediction under the ONNX/TensorRT runtimes specifically (they monkey-patch `forward` to a raw-tensor-only closure) while working fine under the real `nn.Module.forward` (which unwraps dicts internally). Both are fixed to match the pattern already used by the singleview model.

**2. Unified end-to-end inference-speed figure.** Replaces the existing forward-pass-speedup table and the PyNvVideoCodec decode-speed table in `increasing_inference_speed.rst` with a single figure: 3 models (resnet50 singleview, vit-s singleview, vit-s 6-view multiview) x 5 techniques (eager fp32/fp16, torch.compile+fp16, onnx+fp16, tensorrt+fp16) x 2 decoders (DALI, PyNvVideoCodec) x 2 GPUs (L4, A100), 10 repeats per bar with SEM error bars. All numbers are real, genuinely end-to-end `litpose predict` timings, not isolated forward-pass numbers like the old table.

Also includes the fly multiview PyNvVideoCodec-vs-DALI benchmark rerun on 50x-concatenated video (~30k frames/view instead of the original <1000-frame clips), to rule out per-call setup overhead as an explanation for PyNvVC's multiview slowdown.

## Key findings worth a close look

- The old table's "8.32x" speedup claim was from an isolated forward-pass microbenchmark. The real end-to-end number is **2.93x** (TensorRT FP16 + DALI, 6-view multiview, on L4 -- not A100). Smaller because it now includes decode/IO overhead, and L4 shows the biggest relative gain because A100's eager-FP32 baseline is already fast.
- Decoder choice depends heavily on model shape. For single-view models, PyNvVideoCodec is often *faster* than DALI on A100 once any acceleration technique is applied. For the 6-view multiview model, DALI wins outright across every technique/GPU combination -- and the gap is much larger on A100 (44-53% slower for PyNvVC) than L4 (6-19% slower). Confirms the multiview slowdown holds on realistic ~30k-frame videos, not just the original short clips.
- One data gap, documented rather than hidden: L4 hits a genuine VRAM ceiling (22GB) for ONNX Runtime FP16 + DALI on the 6-view multiview model on long video -- fails on the very first predict batch with ~19.6GB already in use before that batch runs. A100 (80GB) has this combination covered. Noted as a caveat in the docs rather than worked around.

## Testing

- Multiview export fix verified via a full predict-speed run across ONNX/TensorRT variants on both L4 and A100 (100/100 rows on A100, 90/100 on L4 with the one documented VRAM-limited combination missing).
- Figure generated directly from the real benchmark CSVs.


lightning_pose/models/heatmap_tracker_multiview.py
themattinthehatt: update to `images: Float[torch.Tensor, "batch views channels image_height image_width"]` and remove the `TypedDict` options

lightning_pose/models/heatmap_tracker_multiview.py
```diff
@@ -335,8 +341,19 @@ def predict_step(
         > predictions = trainer.predict(model, data_loader)
 
         """
+        # Unwrap to a raw images tensor before calling forward(), mirroring
```
themattinthehatt, line 325: great catch!

lightning_pose/models/heatmap_tracker_multiview.py
themattinthehatt: let's streamline the whole model to be consistent with the singleview model; can you update this to:
```
predicted_heatmaps = self.forward(batch_dict["images"])
```

lightning_pose/models/heatmap_tracker_multiview.py
```diff
@@ -234,7 +234,13 @@ def forward(
         """
 
         # extract pixel data from batch
-        if "images" in batch_dict.keys():  # can't do isinstance(o, c) on TypedDicts
+        if isinstance(batch_dict, torch.Tensor):
```
themattinthehatt: let's remove `batch_dict` input entirely and always assume raw image tensor

docs/source/user_guide_advanced/increasing_inference_speed.rst
themattinthehatt: update to ``onnxruntime`` (with double backt-ticks before and after) to match format in point above

docs/source/user_guide_advanced/increasing_inference_speed.rst
```diff
@@ -46,57 +47,60 @@ Overview
   techniques above -- see :ref:`PyNvVideoCodec (Video Decoding) <pynvvc_decoding>`.
```
themattinthehatt: make "Usage: PyNvVideoCodec" like the other options

docs/source/user_guide_advanced/increasing_inference_speed.rst
```diff
@@ -46,57 +47,60 @@ Overview
...
 - **These techniques are complementary and can be combined** -- e.g. torch.compile() or
   TensorRT for the model together with PyNvVideoCodec for decoding -- for the best end-to-end
-  throughput. Not yet benchmarked together on this page; a combined benchmark is planned for a
-  follow-up PR.
+  throughput. The Results figure below crosses every technique with both decoders, so its
```
themattinthehatt: Replace "The Results figure below..." sentence with just "See the figure in the following section."

docs/source/user_guide_advanced/increasing_inference_speed.rst
```diff
@@ -46,57 +47,60 @@ Overview
...
 
-Best-case speedup vs. eager FP32, FP16, isolated forward pass (10 warmup + 100 timed passes,
-largest batch size tested per model/GPU: batch 64 for ResNet50/ViT-S, batch 16 -- effective
-ViT batch 96 across 6 views -- for the multiview transformer):
+The figure below benchmarks every combination of model architecture, precision/runtime
+technique, video decoder, and GPU on real end-to-end ``litpose predict`` calls -- not an
+isolated forward pass. Each bar is the mean of 10 timed repeats (1 discarded warmup run);
+error bars show standard error of the mean.
+
+.. figure:: /_static/predict_speed_matrix.png
```
themattinthehatt: this isn't rendering properly, see here: https://lightning-pose--480.org.readthedocs.build/en/480/source/user_guide_advanced/increasing_inference_speed.html#results

it might be the relative path is incorrect. but either way, I'll ask you to move the figure outside of the repo, just because it makes the repo heavier to install, and if we change/update the figure in the future the old version gets saved forever in the git history, which can itself take up a lot of space eventually. what you can do instead is upload this file to imgur and then copy the link here; here's an example from another section of docs:

```
Drag and drop video file(s) using the provided interface. You will see an upload progress bar.

.. image:: https://imgur.com/GjR2Jb4.png

Choose number of frames to label per video - these frames will be automatically selected to
maximize the diversity of poses from each video.
```

docs/source/user_guide_advanced/increasing_inference_speed.rst
```diff
@@ -46,57 +47,60 @@ Overview
...
+   :alt: Grouped bar chart of end-to-end predict speed across models, techniques, and GPUs
+
+   End-to-end ``litpose predict`` timing for ResNet50 (single-view), ViT-S (single-view), and
+   ViT-S (6-view multiview). Each row shows 5 technique groups (eager FP32, eager FP16,
```
themattinthehatt: multiview -> multi-view

docs/source/_static/predict_speed_matrix.png
themattinthehatt, line 1: update y-axis to frames per second - this will be much more general and informative

docs/source/user_guide_advanced/increasing_inference_speed.rst
```diff
@@ -46,57 +47,60 @@ Overview
...
+  are not a fully strict FP32 baseline either -- doesn't change the direction of any result
+  here, since TensorRT/ONNX Runtime FP16 still win by a wide margin, but it does mean the
+  eager-FP32 bars are a few percent faster than a true FP32 baseline would be for
+  convolution-heavy models like ResNet50.
+- A single dynamic-shape TensorRT profile was used per (model, GPU, precision), covering the
+  full batch-size range tested, rather than a separate engine per exact batch size. This is
+  simpler but can leave some performance on the table at batch sizes far from the profile's
+  optimum -- for example, multiview-FP32-on-L4 was roughly flat (0.91-1.07x) across batch sizes
+  rather than showing a clear win.
+- **ONNX Runtime FP16 + DALI is missing for the 6-view multiview model on L4** -- this
```
themattinthehatt: this seems like a bug in the testing pipeline, the ONNX Runtime session shouldn't be that large and the DALI pipeline takes up basically no memory. one thing to check in your testing suite: do you make sure you dump the CUDA memory after each model test? seems like previous models might still be stored on the GPU.

docs/source/user_guide_advanced/increasing_inference_speed.rst
```diff
@@ -418,10 +422,9 @@ backend that talks to the hardware decoder more directly, skipping DALI's pipeli
 framework overhead.
 
 On a T4, decoding the same video with PyNvVideoCodec measured **8.76x faster** than DALI's GPU
```
themattinthehatt: can remove this paragraph entirely - we'll keep this sort of info contained in the Results section, and these usage sections will be purely about implementation

docs/source/user_guide_advanced/increasing_inference_speed.rst
```diff
@@ -475,70 +478,20 @@ using the backend you asked for.
 Decode speed: L4 and A100, single- and multi-view
```
themattinthehatt: can remove this section entirely - this info is now contained in Results

docs/source/_static/predict_speed_matrix.png
themattinthehatt: see below about moving this image to public hosting site like imgur

themattinthehatt (changes requested):

themattinthehatt (approved):

docs/source/user_guide_advanced/increasing_inference_speed.rst
jjaw123: Good question, and it led somewhere useful — turns out it's not really "previous models still on GPU" in the way I initially assumed either, so I dug into it properly. It's not the ONNX Runtime session (that's freed correctly now — there was a real leak there too, already fixed separately: the eager backbone wasn't being released after attaching the ONNX/TensorRT runtime, and I added arena_extend_strategy to the session's CUDA provider options). The actual cause is DALI itself: it registers every Pipeline object it creates in a process-global list for its own at-exit cleanup, so that object (and its GPU memory pool) stays alive for the life of the process no matter what the caller does with del/gc.collect() — confirmed by finding Pipeline.__del__ and DALI's _shutdown_pipelines() atexit hook directly in the installed package source.

I tried calling the private Pipeline._shutdown() hook manually mid-process to force an early release, and it does free the memory (no more OOM) — but it's not safe: the next pipeline that gets built afterward hits a CUDA-level crash (index out of bounds device-side assert), because that hook is only meant to run once, at interpreter exit, not while the process is still going. So there's no supported way to release DALI's pool early from our code.

Practically: a single, fresh litpose predict call on this exact combination runs cleanly (~479s) — it's only our benchmark script's warmup+repeat loop, which calls predict_video() multiple times in the same process, that runs into it. I rewrote the caveat in the docs to explain this accurately instead of calling it a "memory ceiling" (which wasn't quite right). Getting a real multi-repeat number for this combination would need the benchmark restructured to run each repeat in its own process — happy to take that on as a follow-up if it's worth the effort, but wanted to flag the actual mechanism first rather than guess.

themattinthehatt: thanks for investigating - sounds like you'll need to figure out a way to restructure the benchmark then, we definitely want a complete set of numbers for the docs

themattinthehatt:
@jjaw123 just bumping my last message - if you can quickly update the caveat point then I can merge this PR!

